PR #3, opened 2014-01-26
Moving to proper Python build-system (pip)

SamuelMarks:


Swanson-Hysell:
Moving to a proper build system is a good goal that we haven't yet achieved. The structure of the repository has recently been reconfigured in ways that I think will be helpful for such a goal (there is now a pmagpy folder for the module that has a proper __init__.py file). Because the directory restructuring was so significant and there have been so many other changes to the project since this pull request was put in there isn't a nice way to merge in this pull request. 

We do point people towards using the Anaconda or Enthought Canopy scientific python distributions with which the needed modules come standard (with a couple exceptions like basemap) which I think works for our user base. But being able to pip install instead of manually adding pmagpy to the PYTHONPATH could be valuable. 

I am going to close this pull request given the above, but this could be opened as an issue. Since no one on the project has packaging experience some tips about what is needed for pypi standards would be helpful.

PR #20, opened 2015-12-03
allow plot_di_mean to accept tuple-formatted colors

lfairchild:


Swanson-Hysell:
I will go ahead and merge this commit in for now. It is good to have implemented a mechanism so that the Fischer means in the site class are colored differently. However, I think it may be better to have it colored to the color of the individual plot_di colors rather than to a random color.

PR #38, opened 2016-01-26
added bootstrap and MM1990 reversal test functions; added flexibility to naming requirements in vgp_calc

lfairchild:


moonshoes87:
This still overwrites functionality that I pushed yesterday to Pmag GUI in pmag_menu.py, pmag_menu_dialogs, pmag_results_extract, some tests, and pmag.py.  It looks like my changes in ipmag.py are fine.  Could you merge those other changes in as well?

Swanson-Hysell:
There is also the issue that git sees it as you wholesale deleting and
replacing ipmag.py which means that the individual additions can't be
tracked and it looks like you are simply contributing all of ipmag. I would
recommend pulling your new/modified functions into a text editor, syncing
up the repo and then putting in just those new functions into the ipmag.py
file that is currently in master.

------------------------
Nicholas Swanson-Hysell
Assistant Professor
Department of Earth and Planetary Science
University of California, Berkeley

mobile number: 510.542.4787
http://www.swanson-hysell.org

PR #39, opened 2016-01-26
added bootstrap and MM1990 reversal test functions; added flexibility to naming requirements in vgp_calc

lfairchild:


lfairchild:
Okay, third time's the charm. Sorry for the confusion -- I should have seen that my pull request was modifying more than just ipmag.

ltauxe:
which MM1990 reversal test is this?  if it is Watson’s V, then that should be maybe not MM1990, but maybe the Watson ref or  Tauxe et al., 2010?   if it is the A,B,C thing, that is a silly way to do reversals tests.  Not sure which one you guys are actually doing.

Swanson-Hysell:
These functions that Luke put together are ipmag implementations of these two command line programs:

- revtest.py (this test is an implementation of the Tauxe bootstrap reversal test that can now be done with ipmag.bootstrap_reversal_test)
- revtest_MM1990.py (this test is an implementation of the Watson V common-mean/reversal test that also reports the critical angle as formulated by MM1990that can now be done with ipmag.MM1990_reversal_test)

MM1990 is the Watson V test with the additional formulation of the critical angle which provides a way to obtain a measure of the amount of information available for the test. *Note that one can simply do the Watson's V test using ipmag.watson_common_mean which is the function underlying ipmag.MM1990_reversal_test*. The issue is that if there are only a few observations, or the directions have quite low precision, then there are situations where the null hypothesis of a common mean cannot be rejected using Watson's V test (thereby constituting a "positive" reversal test), but where that result isn't meaningful or robust. 

Consider two populations of directions with quite low precision that have slightly different means, but are far from antipodal from one another:

```
directions1 = ipmag.fishrot(k=1,n=4,Dec=20,Inc=20)
```
```
directions2 = ipmag.fishrot(k=1,n=4,Dec=340,Inc=-20)
```

These means are far from antipodal from one another, but if you draw these populations, take the antipode of one of them and conduct a Watson V test you will get the result that the null hypothesis of a common mean can not be rejected. The MM1990 formulation recasts the V statistic as a critical angle and in this test would argue that the reversal test is indeterminate (not enough information to draw a robust conclusion). This classification makes sense because the data are so bad and the result that they share are common mean is not particularly useful which why it is helpful to have a classification that says the test is indeterminate. These data and there actual (far from antipodal) means as specified in fishrot are plotted below:
download 5

PR #175, opened 2017-04-26
PmagPy in Python3

Caoimhinmg:
Hey all PmagPy developers, to avoid the [sun setting of the scientific python libraries](http://www.python3statement.org/) on Python2 and to future proof PmagPy from the problems this will bring @Swanson-Hysell, @moonshoes87, and I have developed a fork of PmagPy master which we converted to Python3. Making this change will grantee that PmagPy keeps the support it needs from the numerous scientific libraries it relies on as well as by the newer GUI library wxpython Phoenix which while a bit more unstable than the old wxpython is better built and under current development. 

That said this will require that all developers obtain a Python3 environment either using [Anaconda](https://www.continuum.io/downloads), [Enthought](http://blog.enthought.com/canopy/new-year-new-enthought-python-products/), or just [Python](https://www.python.org/downloads/). It should be noted that Canopy does not yet fully support Python3 so those using Mac or Windows are recommended to try Anaconda. Once you've set up your Python3 environment you will need wxpython Phoenix just like in the old PmagPy you can pip install this as follows:

On Mac or Linux:
```bash
 sudo -H pip install --upgrade --pre -f https://wxpython.org/Phoenix/snapshot-builds/ wxPython_Phoenix
```

On Windows (don't forget to replace PATH_TO_PIP with the actual path):
```DOS
C:\PATH_TO_PIP\pip.exe install --upgrade --pre -f https://wxpython.org/Phoenix/snapshot-builds/ wxPython_Phoenix
```

Once you've done that and if necessary added PmagPy to your path and pythonpath just like in Python2, or run the new dev_setup.py script you should be able to run the GUIs and command line programs as usual. **Note:** As of this moment the command line programs should be cross compatible with both Python2 and 3, but all developers will need to continue development in Python3 or there will be serious errors and the GUIs are strictly Python3 only because of the difference between the GUI libraries.

One of the other large things to address is that this change does require that all developers change their code to Python3 code something that should not be overly daunting as there are only a few changes between the two languages. These are documented in the (mostly complete) notebook Py2toPy3.ipynb in data_files/notebooks/ and can also be found all over the internet like [here](http://sebastianraschka.com/Articles/2014_python_2_3_key_diff.html).

So this merge will likely have many conflicts as a lot of development has happened both on this fork and in the PmagPy main repo, so master will be unstable or some time. For those that wish to continue to use the Python2 version of PmagPy while master is in flux there should now be a Python2 branch off PmagPy master. This should cause no change to the binary releases of the GUIs as once frozen it doesn't matter what language we are using, and @moonshoes87 has come up with a clever solution to having pip know which PmagPy to download when you run `pip install pmagpy` to minimize impact on those that are just users.

Let me know what you all think and what happens when you all give it a whorl.


moonshoes87:
@Caoimhinmg, thanks for all the work you've done!  I've pulled all those changes into a local PmagPy branch, so here are the steps you need to follow to try the new code.

1.  Install Anaconda.  You can create an environment with Python 3 or have Python 3 as your default for Anaconda.

2.  `git checkout Caoimhinmg-master`

3.  If you already have PmagPy in your path, you should be ready to go.  Try it out!

njarboe:
Hello Kevin and Lori,

So I was thinking of removing all of my PmagPy Python 2.7 stuff on my mac
(Enthought Python and my current PmagPy git directory) and starting over my
installation from scratch using Anaconda and ending up with the current
master branch installed which is Python 3. Going forward I will use that
branch and make my changes there. Do you see any pitfalls in this approach?
The future standalones will be built with the master branch, not the
Python2 branch, correct? I remember reading about a release
branch(different from master?), but don't see that branch on the PmagPy
github.

Nick

moonshoes87:
Hi Nick,

Your plan sounds like a good one, although you should not need to remove your git PmagPy directory.  You don't even need to uninstall Enthought Python, you just need to install Anaconda Python and give it priority (by commenting out Enthought Python from your .bashrc or .bash_profile).

With Anaconda, you will be able to switch easily between Python 2 and Python 3 if needed.  Likewise, with the developer install, you can switch easily between the Python 2 PmagPy branch and master.  So you will be able to access both pretty simply.  

Having PmagPy in your PATH via the developer install means that your system will access whichever version of a program your git branch is pointing to.  So, to switch between using PmagPy with Python 2 and Python 3, you will need to do just two things.  

Let's say you have Anaconda installed with python 3 as your default Python, and that you have the developer install already.  If you need to switch back to using Python 2.7 and the Python 2 compatible PmagPy, you can follow these steps:

    # create a Python 2 conda environment (one time only)
    conda create python2_env python=python2.7
    # switch to using that python 2 environment
    source activate python2_env    
    # install any needed packages into this environment (one time only)
    # for example:
    conda install numpy
    pip install --upgrade --pre -f https://wxpython.org/Phoenix/snapshot-builds/ wxPython_Phoenix
    # switch to the Python2 branch (must be in your PmagPy git directory)
    git checkout Python2

To go back to your defaults of Python 3 and PmagPy for Python 3:

    source deactivate
    git checkout master

As far as the standalones are concerned, we just created up-to-date versions for all platforms from the Python 2 branch.  Once master is stable again, you are correct, we will start making new releases off of master.  

And as far as the release branch is concerned, that is just a temporary branch that I create when it is time to do a new release.  What this does is freeze development while I (or whomever else) does some testing on that branch to make sure everything is ok, while still allowing other developers to continue making changes to master.  After making a new release, the release branch gets deleted.

Let me know if you have any additional questions!  Hopefully this should be pretty straightforward.

PR #261, opened 2017-11-02
ensure that plot_di plots a list of colors correctly

yuempark:
prior to this commit, if a list of incs with +ve and -ve values was
passed to plot_di along with a list of colors, the vectors would not
plot with the intended colors

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -1429,11 +1429,17 @@ def plot_di(dec=None, inc=None, di_block=None, color='k', marker='o', markersize
             if inc[n] >= 0:
                 X_down.append(XY[0])
                 Y_down.append(XY[1])
-                color_down.append(color[n])
+                if type(color)==list:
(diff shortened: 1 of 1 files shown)


yuempark:
I'm not sure if it is appropriate, but yes - I had a list of poles that I compiled with ages that spanned ~700 My. I converted these poles to expected directions at a given site (using `pmag.vgp_di`), and wanted to plot them all on a single stereonet using a colormap (i.e. mapping the ages of the directions onto a colormap).

(An additional thought): the previous version of `plot_di` produced a stereonet with a list of colors passed as an argument anyway - it was just that it would unexpectedly assign the colors to the wrong vectors if there were both +ve and -ve inclinations. Therefore, since `plot_di` will accept a list of colors and produce a stereonet anyway, I thought that it might as well do it correctly.

Swanson-Hysell:
Sounds good. As long as the single color arguments works it is nice to have list color list functionality as well.

PR #433, opened 2018-09-05
Irm unmixing

botaoxiongyong:
I would like to add the irm unmxing to pmagpy,
However, I am not very familiar with the pull request.

The pyIRM could be indipendent installed and run as a python package. see here https://github.com/botaoxiongyong/pyIRM

I put the pyIRM.py in /PmagPy/programs, and also add one more dependency (the lmfit) into pmagpy_dependencies.txt, and add pyIRM into programs_list.py.

I am not sure whether this is OK now, please check and contact me if anything wrong.


ltauxe:
great i’ll check it out -  on thing is that we have moved away from names like pyIRM and to py_irm, so can we rename your program that?  
and could you also provide and example file (they go in data_files in a directory named py_irm
then we can put an example of how to use it in PmagPy.ipynb   and then more people will use it.  
lisa

botaoxiongyong:
Yes, no problem, I have replaced all the names as py_irm.

I have add the example files under /data_files/py_irm/, including 
    - the input file: py_irm_example.dat, 
    - the out data file: py_irm_exmaple_fit.csv (after click the 'save data' button), 
    - the out figure: py_irm_example.png (after click the 'save figure' button)
    - the gui figure: py_irm_gui.png

ltauxe:
i’ll have a look.  thanks for your efforts. 
lisa

ltauxe:
I tested this and it works very nicely.  We just have to figure out how to make the install process as smooth as possible.  then we can merge it with the rest of PmagPy and add the example to the PmagPy-cli version.  I don't suppose there is any way to get this to work in a Jupyter notebook?

moonshoes87:
Great, I will work on this tomorrow.

ltauxe:
perfect!  thanks!  I’ll try to add it to PmagPy next week.

moonshoes87:
In general this seems to be working well.  

I made a few small changes.  I changed the environment to /usr/bin/env python, which is consistent with the other programs and works across more platforms. I also added a check for the lmfit module so that when a user first runs the program, they will be prompted to install lmfit if they haven't installed it already.  Since this library isn't yet used elsewhere in PmagPy, not all users necessarily will need to install it.

One last detail: can we consider changing py_irm.py to irm_unmix.py?  All programs in PmagPy are Python and so the second name might be more informative.

My changes are in this branch: https://github.com/PmagPy/PmagPy/tree/pr/433

moonshoes87:
If @ltauxe and @botaoxiongyong are ok with the name change, I can do that in https://github.com/PmagPy/PmagPy/tree/pr/433 and merge it in.

ltauxe:
these sound like reasonable changes.  

also there is a very nice example of how to use it in a notebook which we could add to PmagPy.ipynb

 (https://github.com/botaoxiongyong/pyIRM/blob/master/pyIRM/pyIRM.ipynb <https://github.com/botaoxiongyong/pyIRM/blob/master/pyIRM/pyIRM.ipynb>).

let me know when the coast is clear…. 




lisa

Swanson-Hysell:
@botaoxiongyong Great that you are looking to commit this code to PmagPy. Could you succinctly write-up the methodology you are using to make the fits? A brief write-up in this regard would be a nice thing to have accompany the example within the master PmagPy.ipynb notebook. 

Additionally, it would be great to have examples of additionally functionality of the code that you have implemented within the GUI within the example notebook as well. For example, on the project page (https://github.com/botaoxiongyong/pyIRM) you describe how one can manually adjust all the parameters and see the results immediately, and afterwards, 'refit' the data. Showing an example where data are unmixed with and without the assignment of such a prior would be great.

Also, have you considered ways to assign confidence intervals when you are unmixing multiple components? Such confidence intervals are a nice aspect of the MaxUnMix R code (Maxbauer et al. 2016: http://dx.doi.org/10.1016/j.cageo.2016.07.009).

moonshoes87:
I am closing this PR in favor of: https://github.com/PmagPy/PmagPy/pull/436 which has some additional fixes.

PR #437, opened 2018-09-13
add forc diagrams

botaoxiongyong:
the forc_diagram.py can process conventional and irregular forc diagrams, but the remforc is not yet ready, the developping version is on the way https://github.com/botaoxiongyong/pyFORC/tree/forc_gui .

the forc_diagram can be used in command line, and jupyter notebook.

I included two example data file and a jupyter example in /data_files/forc_diagrams. 

The input data format might be verty different due to different software and instruments, so for the forc_diagram.py, it only read data with the line ['    Field         Moment   '] before measured data, or will just skipt the first 35 rows in data file.

in commanline, it should running with data file and smooth factor (SF):
forc_diagram /forc_data_file.txt 3


moonshoes87:
@botaoxiongyong, we haven't gotten a ton of pull requests from people outside of the PmagPy project, so please bear with us as we hone in our reviewing process!

In general this looks good.  I request a few changes:

1.  Please add command-line flags to forc_diagram.py, since this is the general convention in PmagPy.

I.e.:

    forc_diagram -f forc_data_file.txt -s 3

-f is our usual convention for the input file name, you can decide what makes most sense for the smoothing factor (-s, -smooth, etc.)

You should also include a -h option which prints out a docstring with basic information about how to use the program.

See https://github.com/PmagPy/PmagPy/blob/master/programs/conversion_scripts/s_magic.py for an example of how to implement command line options.

2.  Fix 


#/usr/bin/env python

to 

#!/usr/bin/env python


3.

Fix import statement in the notebook:

screen shot 2018-09-13 at 2 39 07 pm

    from programs.forc_diagram import * 

should work.

This should also point to the actual forc example in data_files:

    forc = Forc(fileAdres='../example/MSM33-60-1-d416_2.irforc',SF=3)


4.  These are not absolutely necessary, but I have two suggestions for the GUI:

- when the forc diagram is complete, it should pop up as the front window (in my testing it seems to get stuck behind my Terminal window and I have to search for it)
- it would be nice if typing `ctrl-c <enter>` on the command line would cleanly end the program, instead of having to click on the window to close.

botaoxiongyong:
@moonshoes87 
1,2 fixed

3 It is an exmaple of how to get the FORC data and plot. I am not sure whether the data_file/forc_diagram is a right directory to put the jupyter notebook. Or you may let me know where I should add the jupyter example, I will do the revisions then.

4 The GUI for FORC is not yet ready, since I also would like to include the remFORC. I only provide the conventional and irregular forc here, the window is the plt.show() function to present the result, not GUI. 

you may try the FORC gui here [https://github.com/botaoxiongyong/pyFORC/tree/forc_gui], it is not yet finished and probably takes longer  time, so it is not added here.

Neverthless, I would strongly recommand using Jupyter notebook to process the FORC diagrams. One would prefer to have their own color, axis, label settings, so the forc_diagram.py is mainly aming to provide the FORC data. It is better for people to stylise their own figures.

moonshoes87:
@botaoxiongyong, thanks,  1 & 2 look good.  The one additional change should be the syntax example should say `forc_diagram.py` not `forc_diagram`.

3.  I will move the example into PmagPy.ipynb after we merge in this PR.  Version control is always annoying with the notebook format, so I will deal with that.

4.  I meant the command-line program (not the interactive GUI you are working on), sorry for the confusion.  This seems to be an issue with the MacOS backend for matplotlib. When I run forc_diagram.py, the figure doesn't pop to the front -- instead I have to search for it behind my other windows.  Have you encountered this issue?  The behavior is better if I specify `matplotlib.use('TKAgg')`.  The tkinter backend should work across platforms, and appears to be better on Mac than the default Mac backend.

5.  One additional issue, the encoding line is invalid, and should be changed to:

    #--\*-- coding:utf-8 --\*-- 

Once we resolve these last issues, I will merge this pull request.  Thanks!

botaoxiongyong:
@moonshoes87 
3 Thant will be great, thanks.

4 No, it have not happend to me. I tested on the MacOS (High Sierra, python-3.5.1, matplotlib-2.2.3), everything is fine, and the figure pop up to the front as active. If I add  `matplotlib.use('TKAgg')`, the window will also pop up, but not as active (just behand the Mac console). May I ask in which console you were running the forc_diagram.py?

Neverthless, the tkinter backend do work across platforms, I will add the `matplotlib.use('TKAgg')` then.

5. fixed

moonshoes87:
I was running forc_diagram on OSX High Sierra 10.13.6 in Terminal.  Anaconda Python 3.6.4, matplotlib 2.2.2.  Not sure why the difference, but if TKAgg works for you I think we are set.

This all looks good, and I will merge it.  Thanks for contributing this!

yanyongg:
The forc_diagram.py  didn't work. Both in commend line and jupyter-notebook. warning says 'cannot convert float NaN to integer'. Do anyone can tell what's the problem?
image
image

botaoxiongyong:
working on that. I just tested running on my computer (Ubuntu 16.04, Anaconda 4.6.14), no such problem. 

@yanyongg found his Anaconda was 32bit though on 64bit PC, he is gonna reinstall and try again.

moonshoes87:
I can confirm that forc_diagram.py also works as expected on OSX 10.13.6, Anaconda 4.6.14.

yanyongg:
I reinstall the **_Anaconda3-2019.03-Windows-x86_64_**. However, it warns the environment has not been activated. Only when I manually activate the default environment "base" every time, the python can work. And the **forc_diagram.py** works in this situation. Who knows how to deal with it?
image
image
image

botaoxiongyong:
Hi @yanyongg , it is so. You need setup python environment for python programs running, anaconda do this for you now. Otherwise you may need setup python environment manually, e.g., specify the python version, install every packages you gonna import in your python files. 

You can edit ~/.bashrc file to activate it automatically when you open new terminal, but you won't like it. Since you do not always need it.

Issue #449, opened 2018-10-09
Use cartopy instead of Basemap for the Demag GUI vgp viewer


moonshoes87:

Swanson-Hysell:
https://github.com/PmagPy/PmagPy/issues/292#issuecomment-425591772

Caoimhinmg:
@Swanson-Hysell I can take a look at it this weekend if y'all want. Sorry I've been so out of commission, but I've been meaning to add a map projection option to that dialog for a long time and I need to learn Cartopy anyway. I'll also see if I can work on some of the lingering Demag GUI issues a bit.

Swanson-Hysell:
If you can that would be great. The new ipmag functions using cartopy could be a helpful guide. And there is now a function in ipmag that plot dp/dm that it could make sense to use for plotting site-level means.

ltauxe:
check out the PmagPy notebook and pmagplotlib.plot_map for help on cartopy implementation.

Caoimhinmg:
@Swanson-Hysell and @ltauxe, sounds good I'll take a look this weekend. There is also plenty of screen room on that dialog maybe I'll add a radio button for dp/dm vs. a95 so you can switch between that shouldn't be too hard.

PR #462, opened 2018-10-25
fix some display issues in Demag GUI's VGP viewer

lfairchild:
Following up on #449, I was having a few issues displaying more than a couple of the projections in the VGP viewer. I believe this is fixed with the changes below. It'd be great if someone could test out the modified viewer (@Caoimhinmg , @Swanson-Hysell ?) prior to merging.


Swanson-Hysell:
I gave this a test and it looks to work well. There is the issues with the halo of the A95 around the outside of the plots that remains. But that isn't a big deal I don't think.

Issue #463, opened 2018-10-25
Demag GUI VGP viewer Mercator/Mollweide projections


moonshoes87:
Some projections don't work well with the Demag GUI VGP viewer.  The stereographic projections work, but Mercator and Mollweide don't look very good.

See the McMurdo data:

<img width="1228" alt="screen shot 2018-10-25 at 3 15 38 pm" src="https://user-images.githubusercontent.com/1836578/47533512-da920080-d868-11e8-9767-d6cb24242c7b.png">
<img width="1228" alt="screen shot 2018-10-25 at 3 16 05 pm" src="https://user-images.githubusercontent.com/1836578/47533524-e978b300-d868-11e8-8307-27a24f377940.png">

@Swanson-Hysell or @ltauxe, any ideas about this?

Caoimhinmg:
Oh wow, that's new, and for some reason it doesn't do that on my machine. Regardless I dug into it a bit and it looks like it may have to do with this [cartopy issue](https://github.com/SciTools/cartopy/pull/324) which was linked to by [this one](https://github.com/SciTools/cartopy/issues/624). Our best bet may be to try to verify it is a cartopy bug and that all those that are experiencing the bug have the newest version of cartopy. Then if the problem persists reopen the issue and try to work with them on a solution. In the mean time the north polar stereographic is the most common projection for this kind of analysis anyway.

moonshoes87:
Here is my cartopy version:

    cartopy                   0.16.0           py36h9263bd1_0 

Have you tried upgrading your cartopy?  I'll also try downgrading and see what happens.

Alternatively, we could wait on this a while.  Looks like 0.17 is expected in early November, so several fixes will be added (including some of our issues).

moonshoes87:
Installed the latest (dev) version of cartopy from Github, in which case Mercator works but Mollweide gives a lengthy error:

```
Traceback (most recent call last):
  File "/Users/nebula/Python/PmagPy/dialogs/demag_dialogs.py", line 205, in on_proj_box
    self.plot()
  File "/Users/nebula/Python/PmagPy/dialogs/demag_dialogs.py", line 282, in plot
    dp["vgp_dp"])*111.32, float(dp["vgp_dm"])*111.32, azi, color=dp['color'])
  File "/Users/nebula/Python/PmagPy/pmagpy/ipmag.py", line 2601, in ellipse
    map_axis.plot(X, Y, transform=ccrs.Geodetic(),**kwargs)
[...]

PR #467, opened 2018-10-31
work-around for vgp plotting issue in mollweide

lfairchild:
Catch `ValueError` resulting from irregular VGP data when plotting in Mollweide projection (see issue #463) and issue an error message with the name of data that failed to plot.


moonshoes87:
@lfairchild, this looks good, thanks.  Once tests pass, I will merge it in.

PR #592, opened 2019-08-26
pmagpy/*.py: Partial cleanup of imports.

schwehr:
- Moved python's imports to the top
- Sorted python's imports alphabetically
- Removed commented out imports

I don't agree with the "from builtins" stuff from futurize, but I left those.
I think those should definitely go away when PmagPy drops support for py2.
See:

  https://github.com/PmagPy/PmagPy/commit/32a4f1c2e0edb6b5462325f47d02c6541cc627ea

I did not alter any of the imports of functions or import of `*` that I would normally do.  That seemed to large of a change for a project that I do not know well.  See here for the style guide I usually follow which has some nice things that are different from pep 8:

http://google.github.io/styleguide/pyguide.html#22-imports


Swanson-Hysell:
This clean-up all looks good to be. @moonshoes87 or @ltauxe if you are able to take a look and review the changes before they are merged that would be good to make sure there aren't any downsides of the removal of the commented out imports (i.e. there is a reason they are hanging around).

PR #601, opened 2020-02-05
fix matrix quoted values and cv validation

rminnett:
Here is a proposed fix for validate_upload3 along with a notebook for testing the fix. It requires the "regex" (and not just the "re") package for the zero-length look-ahead regular expressions to match the quoted strings properly. 

validate_upload2 might also need a similar fix and a unit test with data files that include valid and invalid examples of List, Dictionary, and Matrix values would be good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
         - "export DISPLAY=:99.0"  # make the GUI testing work with the host machine display
         #- "sh -e /etc/init.d/xvfb start"
       install:
-        - pip3 install cython pillow pytest pep8 pyshp shapely regex
+        - pip3 install cython pillow pytest pep8 pyshp shapely
         - pip3 install six requests pyepsg owslib cartopy future numpy==1.16.1 scipy
         - pip3 install pandas libtiff
         # get latest version of wxPython4:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 
 matrix:
   include:
-    # works!
     - name: "Python 3.7.1 on Xenial Linux"
       python: "3.7"           # this works for Linux but is ignored on macOS or Windows
       dist: xenial          # required for Python >= 3.7.  BUT doesn't seem to work with conda.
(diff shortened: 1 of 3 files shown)


rminnett:
The Travis CI builds are failing without the regex package.

moonshoes87:
@rminnett, thanks for this.  I haven't used the `re` module extensively, but could you explain what `regex` provides that is missing in `re`?  I'm not sure I understand.

rminnett:
@moonshoes87, the `re` module doesn't implement zero-width matches properly, which are necessary for the look ahead quote matches in this case. https://pypi.org/project/regex/ seems to suggest that in python 3.7+ it does a better job, but when I was testing with 3.8.1 it worked with `regex` and not `re`. Maybe the regular expression I'm using for splitting is an edge case that is still not handled correctly in `re`.

rminnett:
@moonshoes87 In case the regular expression itself is too cryptic, here's an explanation of the matching looking for pairs of quotes in the rest of the string, indicating that the splitting character is outside of a quoted string:
```
;           // Split on semicolon
(?=         // Followed by (zero-width matching)
   (?:      // Start a non-capture group so the result of splitting will be correct
     [^"]*  // 0 or more non-quote characters
     "      // 1 quote
     [^"]*  // 0 or more non-quote characters
     "      // 1 quote
   )*       // 0 or more repetition of non-capture group (multiple of 2 quotes will be even)
   [^"]*    // Finally 0 or more non-quotes
   $        // Till the end  (This is necessary, else every comma will satisfy the condition)
)
```

moonshoes87:
Thanks for walking me through this, Rupert.  Could you give me an example of where this is failing with `re`?  I wasn't able to find an edge case where `re` and `regex` didn't produce the same results.

I was hoping to avoid adding a new dependency if possible, since installing everything is already a challenge for many users.  The Anaconda distribution comes bundled with `re` but not `regex`.  Let me know what you think.

rminnett:
You're right! I just checked on Python 3.7.6 and 3.8.1 and they both work with the `re` module. Perhaps I was testing with an old kernel attached to the notebook at the time! I pushed another commit without the `regex` import.

As I was thinking about it, it might also be good to report an error if there's an odd number of double quotes in the `cell_values` string because the `re.split()` won't handle it correctly and it's probably a mistake.

PR #618, opened 2020-05-18
add functionality to plot poles with filled error ellipses

yuempark:
Can be useful when trying to plot APWP's from multiple cratons at the same time - markers can be colored by age of pole, and error ellipses can be colored by craton.


Swanson-Hysell:
Thanks @yuempark for this addition! I edited it so that the functionality is within the plot_pole and plot_poles() function rather than being separate functions — this should make code maintenance more straightforward.

PR #621, opened 2020-07-24
improve plotting of great circles on orthographic axes

yuempark:
On the default cartopy orthographic axis, when attempting to plot a great circle, the resulting circle would be plotted with such low resolution that it would appear as line segments instead of a circle. Increasing the resolution of the orthographic axis is required to solve the problem, and should not affect standard use of the orthographic axis (as far as I am aware...).

Solution was found here:

https://stackoverflow.com/questions/40270990/cartopy-higher-resolution-for-great-circle-distance-line

### DIFF
--- a/programs/conversion_scripts/run_multi_samples.py
+++ b/programs/conversion_scripts/run_multi_samples.py
@@ -72,17 +72,21 @@ def main():
     site_list=""
     samp_list=""
     speci_list=""
+    image_list=""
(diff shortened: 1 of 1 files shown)


yuempark:
Apologies - the branches in my fork were a little messed up. I have the resolved the issue now, and the changes for this pull request can be found in #622 and #623.

PR #622, opened 2020-09-23
Fix transformation in plot_pole()

yuempark:
When using ipmag.plot_pole() on an orthographic map axis, the following value error was raised by cartopy:

ValueError: Invalid transform: Spherical scatter is not supported - consider using PlateCarree/RotatedPole.

Switching the transformation to PlateCarree seems to have resolved the issue.


Swanson-Hysell (approved):

Swanson-Hysell (commented):
@yuempark Can you explain this LowerThresholdOrthographic() change and why you implemented it?

yuempark:
Ah - that LowerThresholdOrthographic() change was from a previous pull request that I implemented that I guess hasn't been accepted yet:

https://github.com/PmagPy/PmagPy/pull/621

From that previous pull request:

> On the default cartopy orthographic axis, when attempting to plot a great circle, the resulting circle would be plotted with such low resolution that it would appear as line segments instead of a circle. Increasing the resolution of the orthographic axis is required to solve the problem, and should not affect standard use of the orthographic axis (as far as I am aware...).
>
>Solution was found here:
>
>https://stackoverflow.com/questions/40270990/cartopy-higher-resolution-for-great-circle-distance-line

PR #623, opened 2020-09-23
Fix "edge" argument in plot_di()

yuempark:
Perhaps this issue only appears in more recent versions of matplotlib, but passing an empty string as the argument for "edgecolors" in plt.scatter() raises an error.

Therefore replace the default value of "edge" in plot_di() from an empty string ("") to None.


yuempark:
Sorry - not sure why but the commit for https://github.com/PmagPy/PmagPy/pull/621 is included in this pull request...

PR #631, opened 2021-06-29
XPEEM 2 MagIC conversion added

cmaurel37:


njarboe:
Replaces old xpeem converter program and creates function in convert_2_magic.py

PR #650, opened 2022-04-26
add function find_ei_Kent to ipmag

duserzym:
this requires additional matplotlib library colors and cm to be imported in ipmag.py
a new function find_ei_Kent is added
the function allows one to obtain a Kent ellipse to parameterize the output of find_ei algorithm such that uncertainties in inclination shallowing can be represented in pole space

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -460,6 +460,31 @@ def print_pole_mean(mean_dictionary):
     print('Precision parameter (k) estimate: ' +
           str(round(mean_dictionary['k'], 1)))
 
+def print_kent_mean(mean_dictionary):
+    """
+    Does a pretty job printing a Kent mean and associated statistics.
+
+    Parameters
+    ----------
+    mean_dictionary: output dictionary of ipmag.kent_mean
+
+    Examples
+    --------
+    Generate a Kent mean using ``ipmag.kent_mean`` and then print it nicely
+    using ``ipmag.print_kent_mean``
+
+    """
+    print('Plon: ' + str(round(mean_dictionary['dec'], 1)) +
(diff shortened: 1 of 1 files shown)


Swanson-Hysell:
What is the reason to add this code to ```find_ei```?

```
    plt.xlim(min(I)*0.8, max(I)*1.05)
    plt.ylim(min(E)*0.5, max(E)*1.05)
```

duserzym:
> What is the reason to add this code to `find_ei`?
> 
> ```
>     plt.xlim(min(I)*0.8, max(I)*1.05)
>     plt.ylim(min(E)*0.5, max(E)*1.05)
> ```

Without the axes limits the inclination-elongation figure will zoom out and be too small. The intersection between the TK03 model and bootstrap data become messy to view.

duserzym:
Screen Shot 2022-07-19 at 14 57 07
The first picture is a run with 150 synthetic data without the two lines of code
Screen Shot 2022-07-19 at 14 57 56
This second one is with the axes limit constraints.

I realize these two lines with axes limits hard-coded in the function aren't ideal. Is there a better way to do this?

duserzym:
The `tight_axes` optional parameter is added. I didn't quite understand the second comment. Did you mean to put the red curve below the bootstrap curves? It would look like this:
Screen Shot 2022-07-19 at 15 18 29
:

Swanson-Hysell:
I am referring to the black TK03.GAD E/I curve, but this can be an edit down the line as this is a distinct issue to the find_ei_kent() function. Can you change the name of that function so that kent is in lowercase as that is the convention we have throughout the library?

Swanson-Hysell (approved):

PR #654, opened 2022-07-31
add a new ipmag.find_compilation_kent function to ipmag

duserzym:
This function can help generate Kent statistics for a legacy sedimentary paleomagnetic pole where only mean pole lon, lat, A95, and site lon, lat are available. It takes these parameters and an assumed normal distribution for f factor (default is mu=0.62, std=0.14) and generate resamples of mean poles with various inclination corrections and summarize the results with a Kent distribution. Nick and I need to discuss the final values for recommended mu and std. But the function works on the example Cut Face data pretty well:
Screen Shot 2022-07-30 at 23 45 17

ps. we need to allow specifying zorder values in `ipmag.plot_vgp` and `ipmag.plot_pole` functions......!

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -8987,6 +8987,83 @@ def find_ei_kent(data, site_latitude, site_longitude, kent_color='k', nb=1000, s
     else:
         return kent_stats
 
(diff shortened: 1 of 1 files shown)


Swanson-Hysell:
This functionality is not going be incorporated at this point in favor of a bootstrap resampling approach.

PR #658, opened 2022-08-30
Changes in pmag.fisher_mean()

facusapienza21:
I made a few changes to the `pmag.fisher_mean()` functions: 

1. Change the computation of the mean value by using NumPy function instead of native for loops in Python. This shows a speedup of around x2/x3 in computations of the Fisher mean for ~100 points. This is important in simulations where Fisher means are computed multiple times and sometimes calls of the functions are nested. 
2. Added a default return when computing the fisher mean of a sample with one single point. For this case, it returns the inclination and declination of the single sample without computing the rest of the information. 
3. The original function has as argument `di_block`, but the function inside was making calls to a variable called `data`. I am not sure if this was a bug or if it was intentional, but the argument of the function seems not to be used inside the function.  

Since this is a core function of the package, it will be good to perform some tests to know this doesn't break any other part of the library. I made a few tests and it returns the same output that the original function but faster and accounting for exceptions 2 and 3. I tried to run the `python -m unittest discover` command suggested in the [testing README](https://github.com/PmagPy/PmagPy/blob/master/pmagpy_tests/README.md) but it returns errors, which I assume could be just for floating number approximations. For example,

Screen Shot 2022-08-30 at 2 26 50 PM

seems to be comparing float numbers. Maybe by using [numpy.testing](https://numpy.org/doc/stable/reference/routines.testing.html) this won't give this error. 

Happy to continue working and exploring this if this seems to be a useful contribution to the project :)

### DIFF
--- a/pmagpy/pmag.py
+++ b/pmagpy/pmag.py
@@ -4367,13 +4367,13 @@ def fillkeys(Recs):
     return OutRecs, keylist
 
 
-def fisher_mean(di_block):
+def fisher_mean(data):
     """
     Calculates the Fisher mean and associated parameter from a di_block.
 
     Parameters
     ----------
-    di_block : nested list of [dec,inc] or [dec,inc,intensity]
+    data : nested list of [dec,inc] or [dec,inc,intensity]
 
     Returns
     -------
@@ -4407,7 +4407,7 @@ def fisher_mean(di_block):
     X = np.array(dir2cart(di_block))
     Xbar = X.sum(axis=0)
     R = np.linalg.norm(Xbar)
-    Xbar /= R
+    Xbar = Xbar/R
     dir = cart2dir(Xbar)
 
     fpars["dec"] = dir[0]
--- a/pmagpy/pmag.py
+++ b/pmagpy/pmag.py
@@ -4398,13 +4398,13 @@ def fisher_mean(data):
      'alpha95': 6.414731246264079,
      'csd': 4.20876891770567}    
     """
-    N, fpars = len(di_block), {}
+    N, fpars = len(data), {}
     
     if N < 2: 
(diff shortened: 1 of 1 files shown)


Swanson-Hysell:
Thanks for making this pull request. 

The bug that you found regarding the name of the input variable was introduced in this commit. 
https://github.com/PmagPy/PmagPy/commit/570989520b74eb3142acab57efbae2d474c03824
This commit was from a student improving docstrings who looks to have inadvertently introduced this bug (which has since been fixed). It has been been fixed back to data:
https://github.com/PmagPy/PmagPy/commit/8c5080f07158a6b70962fefc475b022956c31976

I used `git blame` and saw that before this commit, the line had been untouched and had been ```def fisher_mean(data):``` for the past 12 years.

I just made a couple of commits to revert your branch so that it is using data. I also changed /= to be an explicit division as that is more readable for those that more deeply versed in Python.

Swanson-Hysell:
It would be really great to your contributions and improvements to the tests @facusapienza21. We have lagged on this front since @moonshoes87 moved on from the project. I think that the issue you point to can be solved using ```self.assertAlmostEqual()```

facusapienza21:
Excellent! Thank you @Swanson-Hysell for incorporating the changes! 

One short question: is it recommended to use explicit division instead of `/=`? My first guess is that `/=` is more readable, but I may be missing something. Thank you!

Swanson-Hysell:
My thought is that someone who doesn't know Python can read and understand explicit division where as they would need to understand the augmented assignment operators in order to read `/=`. So I think it makes the code more accessible to have explicit division, even though the augmented assignment operators are more concise (and a tiny bit faster). I could be wrong here and am happy to revisit.

It would be great to interact more about the tests — I have more to learn in this realm.

PR #659, opened 2022-10-05
expose zorder parameter for plot_pole and plot_vgp functions

duserzym:
I edited `ipmag.plot_pol`, `ipmag.plot_poles`, and `ipmag.plot_vgp` functions so that they now have a new optional parameter `zorder` which allows users to define the pole or vgp layer order when rendering the plot.


Swanson-Hysell (commented):
update default zorder value

PR #660, opened 2022-10-05
update docstring for plot_pole

duserzym:


Swanson-Hysell (commented):
add optional zorder specification to plot_vgp() docstring as well

Swanson-Hysell (approved):
I updated the docstrings to include the default zorder info

PR #664, opened 2022-10-13
delete ipmag.plot_XY() as not used anywhere

Swanson-Hysell:
The function ipmag.plot_XY is a very limited wrapper around ```plt.plot()``` that is not used anywhere in our codebase nor anywhere else that I can find. I think it makes sense to delete from ipmag.py

```
def plot_XY(X=None, Y=None, sym='ro'):
    plt.plot(X, Y, sym)
```

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -1926,10 +1926,6 @@ def plot_net(fignum=None):
     plt.axis((-1.05, 1.05, -1.05, 1.05))
 
 
-def plot_XY(X=None, Y=None, sym='ro'):
-    plt.plot(X, Y, sym)
-
-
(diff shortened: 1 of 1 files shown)


Swanson-Hysell:
If you could have a quick look at this @ltauxe that would be great. I think it is pretty straightforward to just remove this function, but thought that I would touch base first before doing so. Note that this function that is removed in this pull request is distinct from pmagplotlib.plot_XY() which is used within ipmag.curie().

ltauxe:
yeah i don’t think we need it anymore  - it was a replacement for bob parkers old fortran code for plotxy.  
lisa

Swanson-Hysell:
Thanks. I just merged the pull request with the deletion.

PR #671, opened 2022-11-10
Add Bogue+Coe 1981 simul algo

apivarunas:
This adds the Bogue and Coe (1981) simultaneous algorithm functionality. This allows users to check the probability of correlating two directions with known precision parameter estimates and a known angle between them. This was written in Python by Scott Bogue, and I translated it into PmagPy functionality to hopefully make it more widely available.

This is my first time doing a commit. I'd like to start helping with this project more in the future.

Bogue, S.W., and Coe, R.S., 1981, Paleomagnetic correlation of Columbia River basalt flows using secular variation. Journal of Geophysical Research, v. 86, p. 11883–11897.

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -15257,3 +15257,42 @@ def validate_magic(top_dir,doi=False,private_key=False,contribution_id=False):
 
 
     return magic_dir,upload_file
+
+def prob_simul(alpha,k1,k2,wordy=False):
+    """
+    the function ipmag.prob_simul runs the Bogue + Coe (1981) algorithm for probabilistic correlation. k1 and k2 can be naively estimated from kappa
+    or one can use the function ipmag.full_kappa to get a better estimate as in the original publication
+    
+    Parameters
+    ----------
+    
+    alpha: angle between paleomagnetic directions (site means)
+    k1: kappa estimate for first direction
+    k2: kappa estimate for second direction
+    
+    Returns
+    --------------------------
+    
+    simul_prob
+    """
+    trials=10000
+
+    hit=0
+    miss=0
+
+    for i in range(trials):
+        #FakeFish returns a list of [dec,inc]s, in this case just 1.
+        lontp1,lattp1 = ipmag.fishrot(k1,1,0,90,di_block=False)
+        lontp2,lattp2 = ipmag.fishrot(k2,1,0,90,di_block=False)
+        angle=pmag.angle([lontp1[0],lattp1[0]],[lontp2[0],lattp2[0]])
+        if (angle>=alpha):
+            hit=hit+1
+        else:
+            miss=miss+1
+
+    p1=1.0*hit/trials
+    simul_prob=p1;
+    
+    if wordy == True:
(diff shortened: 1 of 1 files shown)


Swanson-Hysell (changes requested):
Great to have this contribution. A couple tweaks that would be good before merging:

1. I think that the title of the function could be more informative than prob_simul(). Perhaps strat_correlation_prob() or something along those lines?
2. It makes change the "wordy" parameter to be "print_result" rather than "wordy" as we use that as the parameter name for similar functionality elsewhere. Also, go ahead and put it in the parameter list. Typically the docstrings note what the default parameter is (in this case True). 
3. It makes sense to have trials be a changeable parameter as well. You can put it into the function with the default value of =10000

Swanson-Hysell:
@apivarunas Great to have your contribution and it would be fantastic to have more contributions to PmagPy from you moving forward. I made a couple of suggestions in the code review that can help standardize things a bit in terms of parameter names and the docstring.

apivarunas:
Thanks for the comments Nick. I've addressed them as follows:

1. named the function 'ipmag.simul_correlation_prob'. This is more informative, while creating the natural next name for the BC81 'rand_correlation_prob' algorithm which is just down the pipeline.
2. I renamed 'wordy' to 'print_result' and added to docstrings
3. I added trials to the parameter list with that default value and changeability.

Swanson-Hysell (approved):

Swanson-Hysell:
Good deal. Could you also add an informative description to the docstring about what the output probability parameter is?

Swanson-Hysell:
One other small change is that you have this line:
```
simul_prob=p1;
```
put that defined variable ```simul_prob``` is not used as you return ```p1``` which means that line could be deleted. Or alternatively you could return simul_prob or change the line:
```
 p1=1.0*hit/num_trials
```
to
```
simul_prob=1.0*hit/num_trials
```
and return ```simul_prob```

Similarly it seems like you should just have the parameter be ```num_trials``` rather than having this line in the function:
```
num_trials=trials
```

All small things, but implement how you see fit and we can then merge the pull request.

apivarunas:
Thanks for that attention, Nick. I have added to the docstring, which should now make it clear what is being returned and how to find more information. I also added reference to the potential companion functions, which I can add in a new branch, or just keep working in this branch if that makes sense. Let me know what y'all prefer!

The redundant variables are now fixed, I decided to stick with `simul_prob` and `trials`, but now they don't get transmuted into redundant variables.

The `print result` output was giving both the explanatory sentence with `simul_prob` incorporated, and the returned value itself, so I made it only return `simul_prob` if `print_result=False`

Swanson-Hysell (approved):

Swanson-Hysell:
Great. I just merged in the pull request. I think putting in future pull requests from this branch or from another one should both work fine.

One thing that we also have done in some of the docstrings that can be nice is to put in an example using the function — you can look at the formatting for those for some other functions. One thing that is nice about doing is is that they then get nicely rendered in the documentation: https://pmagpy.github.io/PmagPy-docs/api/api.html#pmagpy-ipmag

PR #680, opened 2023-02-06
added the 'random' probability test of Bogue and Coe 1981

apivarunas:
Have now added the 'random' algorithm of Bogue and Coe (1981). This tests the alternative scenario to the 'simul' algorithm (i.e., the probability of the directions being acquired at different times).

function is named 'rand_correlation_prob' in order to mirror 'simul_correlation_prob'

Next plan is to improve docstrings for both with examples.


Swanson-Hysell (commented):
It makes sense to return rand_prob whether or not the print_result=True. I made a commit to accomplish this.

PR #685, opened 2023-03-04
removed some old_div from pmagplotlib re issue#600

apivarunas:


duserzym:
Thanks for doing this Anthony! I looked through the changes and spotted one place where old parentheses were left undeleted when removing old_div functions. Otherwise all looks ok to me. @Swanson-Hysell could you give this a look through before merging? There aren't so many old_divs in pmagplotlib. It will be good to have one more pair of eyes on the changes just in case.

Swanson-Hysell:
Thanks for popping in and getting rid of these `old_div()`s @apivarunas! I pulled off a couple more stray parentheses and will got ahead and merge the pull request.

PR #687, opened 2023-04-01
Efficient implementation of Fisher samples + Test suite using PyTest

facusapienza21:
This PR implements:
- Changes in the function `fishrot` used to sample from a Fisher distribution. See this [issue](https://github.com/PolarWandering/PaleoSampling/issues/9) for more information about changes. 
- Test suite using Pytest. The folder `pmagpy/test` has a series of tests that can be automatically executed by entering `pytest` in a terminal inside the folder `pmagpy`. We can also automatize this by adding it to CI with GH Actions (see [here](https://ucb-stat-159-s23.github.io/site/lectures/testing/workflow.html) for information of how to do this).

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -488,7 +488,7 @@ def print_kent_mean(mean_dictionary):
 def fishrot(k=20, n=100, dec=0, inc=90, di_block=True):
     """
     Generates Fisher distributed unit vectors from a specified distribution
-    using the pmag.py function pmag.fshdev() and pmag.dodirot() functions.
+    using the pmag.py function pmag.fshdev() and pmag.dodirot_V() functions.
 
     Parameters:
         k : kappa precision parameter (default is 20)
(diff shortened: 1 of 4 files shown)


Swanson-Hysell (commented):
Everything looks good here in terms of output from the updated ipmag.fishrot() and the tests are running well. Should we update pmag.fishdev() and this function so that we can pass a seed in order to have reproducible samples generated?

ltauxe:
yes lets

facusapienza21:
I think fixing the seed is good but just for the test, to ensure the results are indeed reproducible. I would recommend not fixing the seed in the function itself since this can lead to the opposite problem: having users sample the same sequence every single time (this is also a problem when we parallelized the code). Adding an optional argument to `fishdev()` could be an option too.

Swanson-Hysell:
Yes. I see this not has having the default be to fix the seed, but rather to allow setting the seed to be an option.

PR #688, opened 2023-04-01
add zorder optional parameter to plot_di and plot_di_mean

duserzym:


Swanson-Hysell:
Adding this as an optional parameter is a good idea. However, does it make sense to set the default to be 1? What typically happens when a zorder is not set in matplotlib? Do we want to be overriding that with 1?

Swanson-Hysell:
I guess that doing the same as line (zorder = 2) might make the most sense?

duserzym:
I just made the changes for the default zorder values.

duserzym:
I keep running into this zorder issue when doing stereonet plots. I am merging this branch to main.

Swanson-Hysell:
Good deal!

Issue #692, opened 2023-06-16
Polar VGP α-95 ellipses have artefacts


aluthfian:
**What command was run:**

1. Loading the paleomagnetic inclination, declination, and $\alpha_{95}$ from the csv file using pd.read_csv
2. Calculating the VGP using `ipmag.pmag.dia_vgp`
3. Make an orthographic basemap using `ipmag.make_orthographic_map`
4. Plot the VGP iteratively using `ipmag.plot_pole_dp_dm` in a `for` loop
5. This map with the real ellipse and artefact lines is produced.

**What happened:**

<img src="https://github-production-user-asset-6210df.s3.amazonaws.com/47322232/246295878-ce79ee18-1d33-4085-b242-fa37523f9bab.png" width="400">

**What I expected to happen:**

The same globe image as above but without choppy polygon lines near the pole.
<img src="https://github-production-user-asset-6210df.s3.amazonaws.com/47322232/246302107-e2931ecf-f565-4ee3-8f8b-520983d4ad41.png" width="400">

**Data file(s):**

N/A (closed data)

**PmagPy version:**

_pmagpy-4.2.109_

**My workaround**

To define transform in the `ipmag.plot_pole_dp_dm` function. Works with `ccrs.Geodetic()`.

```
def plot_pole_dp_dm(map_axis, plon, plat, slon, slat, dp, dm, pole_label='pole', site_label='site',
                    pole_color='k', pole_edgecolor='k', pole_marker='o',
                    site_color='r', site_edgecolor='r', site_marker='s',
                    markersize=20, legend=True, transform=ccrs.PlateCarree()):

    if not has_cartopy:
        print('-W- cartopy must be installed to run ipmag.plot_pole_dp_dm')
        return
    dp_km = dp*111.32
    dm_km = dm*111.32
    map_axis.scatter(plon, plat, marker=pole_marker,
                     color=pole_color, edgecolors=pole_edgecolor, s=markersize,
                     label=pole_label, zorder=101, transform=transform)
    map_axis.scatter(slon, slat, marker=site_marker,
                     color=site_color, edgecolors=site_edgecolor, s=markersize,
                     label=site_label, zorder=101, transform=transform)
    # the orientation of the ellipse needs to be determined using the
    # two laws of cosines for spherical triangles where the triangle is
    # A: site, B: north pole, C: paleomagnetic pole (see Fig. A.2 of Butler)

    site_lon_rad = np.deg2rad(slon)
[...]

PR #693, opened 2023-06-16
add transform optional parameter to function plot_pole_dp_dm

duserzym:
I hope to make this change to address issue #692


Swanson-Hysell:
Should we make the default ccrs.geodetic() or will that cause other issues?

duserzym:
My experience is that PlateCarree usually works fine and geodetic has other issues such as making weird filled polygons instead of a circle/ellipse, and I felt that geodetic more often has issues for our applications.

Swanson-Hysell:
That makes sense. Perhaps we can add to the documentation of the parameter that it will rarely need to be changed?

Swanson-Hysell:
This should do the trick!

PR #703, opened 2023-09-25
integrate functions to apply Heslop et al. (2023) bootstrap confidence region approach

Swanson-Hysell:
This commit changes the pmag.form_Mhat function to be restricted to the way the function was published in the Heslop et al. (2023) notebook. It also develops ipmag.py functions that use these pmag functions to conduct the test and plot the results. Examples of the code in use are within this notebook: https://pmagpy.github.io/PmagPy-docs/example_notebooks/template_notebooks/mean_determination.html

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -381,6 +381,69 @@ def kent_distribution_95(dec=None, inc=None, di_block=None):
         return pmag.dokent(di_block, len(di_block), distribution_95=True)
     else:
         return pmag.dokent(di_block, len(di_block), distribution_95=True)
+    
+    
+def mean_bootstrap_confidence(dec=None,inc=None,di_block=None,num_sims=10000,alpha=0.05):
+    """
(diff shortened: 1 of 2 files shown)


Swanson-Hysell:
@ltauxe I am going to go ahead and merge this pull request as I now have an associated example notebook up on https://pmagpy.github.io/PmagPy-docs/example_notebooks/template_notebooks/mean_determination.html but submitted it here as a pull request rather than a direct commit in order to group these two commits. Note that I changed your `pmag.form_Mhat` function to not calculate the mean within it as that seemed to be mixing the directional array and the mean that it wanted. I think it is all working as it should be now and the ipmag functions should make it convenient for folks to use. @dave-heslop74 Perhaps you can look at this implementation too. It wraps the notebook code in a function that accepts dec, inc lists or a di_block. I put it to use on the same McFadden et al. (1990) data here: https://pmagpy.github.io/PmagPy-docs/example_notebooks/template_notebooks/mean_determination.html

Issue #706, opened 2023-10-09
ipmag.unpack_magic/ipmag.download_magic not parsing MPMS data


Swanson-Hysell:
**What command was run:**

```
ipmag.unpack_magic('./MPMS_data/ECMB 2018.txt',dir_path='MPMS_data')
```

On this file:

[ECMB 2018.TXT](https://github.com/PmagPy/PmagPy/files/12850032/ECMB.2018.TXT)

**What happened:**

Successfully unpacked hysteresis experiments, but wasn't able to parse the MPMS data giving this message:

```
working on:  'measurements\r'
WARNING:  problem in file with line: 
meas_temp	meas_field_dc	magn_mass	magn_r2_det	timestamp
skipping....
WARNING:  problem in file with line: 
10.01318	0	0.0188512354679803	0.9812937	2020:01:06:21:12:37.36 
```

We need to diagnose the issue and either figure out if there is something wrong with the IRM output or with the ipmag.download_magic function.

duserzym:
This is error comes from the mismatch in the number of columns with the expected. 

Because `ipmag.download_magic' determines the expected column number and format with the typical magic files in mind. One example is below. It basically looks at the row below 'tab site' for the number of columns and the column names ('keys' in the image below), and check that each entry has the same number of columns when unpacking the data file. 
![Screenshot 2023-10-11 at 09 35 14](https://github.com/PmagPy/PmagPy/assets/39976081/81ad7ef5-2484-4351-9a97-e19708d3c073)

However, in the provided file (IRM data written into pseudo-MagIC format), the first line below 'tab measurement' is specimen and specimen name, which would be interpreted as indicating there being only two columns in the measurement file. But the hysteresis loop measurements have three and the MPMS data have 5. The reason that the hysteresis files got unpack with no apparent problem is because it was dealt with as an exception but for the wrong reason. 
![Screenshot 2023-10-11 at 09 40 32](https://github.com/PmagPy/PmagPy/assets/39976081/3eb6f570-98f2-42e3-9297-78f3b8b8266a)

Overall, I think this is a formatting issue associated with writing IRM data files into MagIC format. One easy way to resolve the issue is to have all the IRM metadata (those lines that start with *) be appended at the end of the file instead of being at the beginning. I attach a short modified file that works now. 

[ECMB_2018.TXT](https://github.com/PmagPy/PmagPy/files/12872572/ECMB_2018.TXT)

Swanson-Hysell:
[...]

PR #707, opened 2023-10-21
update download_magic function to fix unpacking IRM rock mag files

duserzym:
update to the ipmag.download_magic function in response to issue #706


Swanson-Hysell:
Looks good. In the future, we might want to figure out a way to name the measurement files in an informative way and/or generate a measurement summary table given that it is a bit daunting to see a long list of measurement files for each experiment with no context for them. 

Screenshot 2023-10-30 at 10 33 15 AM

Issue #697, opened 2023-07-03
doincfish function bug


xiaoneimeng:
The doincfish function cannot be used properly due to the fact that ML_zeros = np.array(Oo[[idx_zeros]]) constructs ML_zeros as a three-dimensional array. This issue can be resolved by modifying the line to ML_zeros = np.array(Oo[idx_zeros]).
Additionally, after making the modifications and running the example file again, the calculated α95 in the output results differs from the provided results. However, the example's expected output described in the function remains the same. I am currently reviewing the specific calculations.

**PmagPy version:**
4.2.109

Swanson-Hysell:
@ltauxe Are you able to take a look at this issue regarding `pmag.doincfish`?

apivarunas:
See discussion in #713, @Swanson-Hysell 

I reproduced an error in the doincfish function and fixed it. I could not reproduce the change in a95. So, the function works now. I made some additional changes in that PR, as well.

Swanson-Hysell:
Thanks for raising this issue @xiaoneimeng. We should be all good now thanks to @apivarunas and #713

PR #713, opened 2024-06-12
In reference to #697

apivarunas:
I examined the function doincfish in the context of #697. Running the function as it existed (double brackets in line 6325) I encountered an error message ("ValueError: cannot reshape array of size 2 into shape (1,1)"). I removed the additional brackets as suggested and ran it. The function worked, and the calculated a95 was 57.45300272498923 versus the docstring value of 57.453002724988956. In other words, the function seems to be working as intended.

While I was at it, I changed the confidence interval calculations to be asymmetric about the mean inclination, removed some old_div, and rounded off the returned values.

Sorry to have included multiple things in the same pull request.


Swanson-Hysell (changes requested):
Thanks for digging into this. A couple changes that I think make sense before merging:
- It is great that you took out the old_div() as we need to be doing this everywhere. There isn't a need to have a comment on the line of code where this is taken out.
- I don't think it is best practice to be rounding the values that are returned. Doing it to 5 decimal point makes it so that it shouldn't be a problem, but the only reason to do the rounding is for the purposes of display being cleaner. There are use cases where getting rid of decimal points is a problem (say for anisotropy data) so I think that we should not modify the floats that are returned from the functions for the purposes of a cleaner display. Instead, we can make convenience functions that display dictionaries like we have done for the Fisher mean dictionaries. So I would request getting rid of the np.round() functions.

Swanson-Hysell:
@ltauxe It is great that @apivarunas dug into the code to make this pull request that addresses issue #697. Can you have a look at it before we move forward with merging?

apivarunas:
I've made the requested changes, to remove the unnecessary comments as well as to not round off values. Then I needed to change the example to reflect that.

Swanson-Hysell:
@apivarunas What is the method that leads to the asymmetric confidence intervals rather than the symmetric one that is what has been implemented in the function? The method previously implemented in pmag.doincfish() is that of McFadden and Reid (1982) where:
$\cos \alpha = 1 - \frac{1}{2} \left( \frac{S}{C} \right)^2 - \frac{f(N - C)}{C(N - 1)}$
This is the method that Lisa details in: https://earthref.org/MagIC/books/Tauxe/Essentials/#x1-241000C.2.3
and that is also detailed in: http://pmagpy.github.io/PmagPy_calculations.html
so with an update to asymmetric confidence intervals we will want to update documentation and put in associated citations.

Swanson-Hysell:
I think that a solution here is to still calculate and return `alpha95` in the dictionary as that is `alpha95` as defined by McElhinny and McFadden (2000) and McFadden and Reid (1982). Still having it in the dictionary will make it so there are not any backward compatibility issues. We can then update the function to also give the confidence limits. I think that it makes sense to have it be that the confidence limits are as provided in McElhinny and McFadden (2000) where they are returned as the minimum and maximum inclination values (rather than that angle). Updated code building on yours could then look like:
```
    f = pmag.fcalc(2, N - 1)  # the 'g' of MM2000
    a95 = np.rad2deg(np.arccos(1. - (0.5) * (S / C) ** 2 - (f * (N - C)) / (C * (N - 1))))
    lower_confidence_limit = Imle[0] + ((180 * S) / (np.pi * C)) - a95
    upper_confidence_limit = Imle[0] + ((180 * S) / (np.pi * C)) + a95
    csd = 81. / np.sqrt(k)

    fpars["upper_confidence_limit"] = upper_confidence_limit
    fpars["lower_confidence_limit"] = lower_confidence_limit
    fpars["alpha95"] = a95
    fpars["csd"] = csd
    return fpars
```
This code also fixes the weirdness that `a95` was defined twice (once without the `np.rad2deg(np.arccos())` and one with)

Swanson-Hysell:
I think that it would also be good to make the example be a bit better with values that are more akin to ones that someone would see and with smaller resulting uncertainty on the mean. Something like:
```
pmag.doincfish([60, 62, 59, 67, 58, 68, 47])
```

Swanson-Hysell:
Also, do you know where the 30º cutoff is coming for:
```
'WARNING: mean inc < 30, returning gaussian mean'
```

apivarunas:
1. Your solution is good re: updating but keeping alpha95. Code looks good (and certainly looks code), I'll implement it in a commit.
2. I thought something similar myself. I can make that change. Perhaps we can just use the values in the example from McFadden and Reid, 1982 [62.4, 61.6, 50.2, 65.2, 53.2, 61.4, 74.0, 60.0, 52.6 and 71.8]. Gives it a through-line.
3. Not sure _exactly_ where that is coming from. But I believe that the arithmetic mean is less problematic for inclination-only data given relatively shallow inclinations (maybe 30 degrees is a cutoff for that). That's pretty dependent on the estimate of kappa, however. See, for instance, Enkin and Watson (1996)

> Through numerical simulations, we have determined that there is no advantage to performing more complicated estimates of the inclination when theta * sqrt (k*) > 400°. For example, if the mean inclination is 30°, this simple estimate is acceptable as long as k* is greater than 44. But if theta =60° then the result will be biased shallow unless k* > 178, which is seldom the case in palaeomagnetic studies.

Okay, looks like my plans for this Friday night are sorted.

Swanson-Hysell:
Using the example values from McFadden and Reid (1982) is a nice idea. Provides a nice little check of the values that the function is returning as well (and my code looks code typo is pretty funny 😆)

apivarunas:
Okay, I've made the changes as discussed. alpha95 is back in the dictionary, along with the upper and lower confidence expressed as the inclination. Example in docstring is the inclination values from McFadden and Reid (1982).

I tested it versus the original function (with only the brackets removed so that it could run) and get the same values, so it is working as it did before.

new_function_ex
old_function_ex

Swanson-Hysell:
Excellent. Great to have your contribution here!

Issue #718, opened 2024-09-18
Incorrect Handling of Intensity in fisher_mean Calculation


MengqingWangCUG:
**What command was run:**

data = [[-45,150,30],[-40,150,35],[-38,145,40]]
pars=pmag.fisher_mean(data = data) 

**What happened:**
a95 missing

The current implementation of fisher_mean includes data with intensity values, which should not be the case. Including intensity leads to an issue where the a95 value cannot be calculated properly. This is because the calculation of a95 implicitly assumes that the input consists of unit vectors.

In the a95 calculation, the number of vectors n is used to offset the resultant vector length. When intensity values are included, the equation a = 1 - b * (N - R) / R results in a being greater than 1, making it impossible to compute a95.

**PmagPy version:**
4.2.124

duserzym:
Thanks for raising this issue. 

Looking at the `pmag.fisher_mean` function code, the issue is caused by the individual vectors not being normalized when a given data block contains intensity values. 

The related code is:
`X = np.array(dir2cart(data))`
`Xbar = X.sum(axis=0)`
`R = np.linalg.norm(Xbar)`
`Xbar = Xbar/R`
`dir = cart2dir(Xbar)`

@Swanson-Hysell @ltauxe Shall we 1) force normalizing vectors and support intensity input, or 2) force users to only give [dec, inc] values? My opinion is that it probably does not make sense to include intensity values when calculating fisher mean statistics, but in case there are other functions in PmagPy that use this function in the [dec, inc, int] format, perhaps the best solution now is to force vector normalization?

Swanson-Hysell:
Thanks for raising this issue. The first thing it made me notice is that the docstring for pmag.fisher_mean() had a poor example.

This example data is not sensical for what should be a nested list (di_block) of declination, inclination pairs:
```
    >>> data = [[-45,150],[-40,150],[-38,145]]
```

I have just made a commit ([9a458c3](https://github.com/PmagPy/PmagPy/commit/9a458c3b919cb175ff1195fe99795bf645fd3b93)) that updates the example to be more sensical:
```
>>> data = [[150,-45],[151,-46],[145,-38],[146,-41]
```

Regarding intensity values, as @duserzym writes, a Fisher mean is a mean of unit vectors. So I think the approach here is to allow a nested list of [dec,inc,intensity] to be passed to the function, but to then make them unit vectors when calculating and returning the mean.

MengqingWangCUG:
Thanks for following up

PR #719, opened 2024-09-18
update pmag.fisher_mean so that we only work with unit vectors here

duserzym:
address issue #718

### DIFF
--- a/pmagpy/pmag.py
+++ b/pmagpy/pmag.py
@@ -4521,8 +4521,8 @@ def fisher_mean(data):
         return {'dec': data[0][0], 
                 'inc': data[0][1]}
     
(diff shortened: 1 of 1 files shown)


Swanson-Hysell (approved):
This commit looks good. I was going to implement this line as the first line of the function:
```
data = [[d[0], d[1]] for d in data] if len(data[0]) == 2 else [[d[0], d[1], 1] for d in data]
```
but your solution works just as well.

duserzym:
> This commit looks good. I was going to implement this line as the first line of the function:
> 
> ```
> data = [[d[0], d[1]] for d in data] if len(data[0]) == 2 else [[d[0], d[1], 1] for d in data]
> ```
> 
> but your solution works just as well.

PR #721, opened 2024-09-27
Bug in dealing with files without dec_corr assigned

matdomeier:
For 2G binary files where dec_corr is not provided, the present behavior (modified in 2021) skips the azimuth value and assigns the plunge value to the azimuth variable.


matdomeier:
It actually seems that _2g_bin expects the dec_corr value in a different place than where it is reported in our 2G binary files, so rather than modifying this one perhaps it is better if I just make a new function for 'Oslo'. I will scrap this pull request and submit another one with these alternative changes.

Swanson-Hysell:
Thanks for checking this. I think there are different iterations of these 2G binary files. 

Rather than a new function, perhaps the way to go would be make it so that there are different possible options within the `_2g_bin` function that deals with these format difference — such as a keyword argument for dec_corr_position with Oslo being an option?

PR #722, opened 2024-12-31
Fixed some issues with running demag_gui as a program

apivarunas:
After re-installing Pmagpy on a new work computer, I had some issues running demag_gui from the command line as I usually do.

1. At first, the GUI simply did not open at all. I eventually tracked this down to an initial line in thellier_gui_dialogs.py, which was trying to import arange from scipy, whereas arange is in numpy. So, that's fixed now.

2. Next, there was an issue with a number of lines in demag_gui.py:

Various "sizer" commands where there were floats. Got a "TypeError: Sizer.Add(): arguments did not match any overloaded call". Apparently, the arguments wanted ints. Perhaps at some point there was silent conversion of floats to ints going on. Either way, adjusting the floats to the int equivalents (0.5 to 0, 1.5 to 1, etc) fixed this problem.

When adjusting fits, the plots were not updating. Although they'd eventually update with the correct bounds if one moved off a specimen then back, the "Interpretations Editor" was unable to be opened if one had already made fits. The error was: "ArtistList object has no attribute 'remove' " Adjust the .remove() command to act directly on the objects worked a treat.

### DIFF
--- a/dialogs/thellier_gui_dialogs.py
+++ b/dialogs/thellier_gui_dialogs.py
@@ -29,7 +29,7 @@ import wx.lib.scrolledpanel
 import copy
 import os
 import scipy
-from scipy import arange
+from numpy import arange
 # only this one is nessesary.
 import wx
 from matplotlib.backends.backend_wxagg import FigureCanvasWxAgg as FigCanvas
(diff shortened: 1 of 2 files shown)


duserzym:
firing up thellier_gui.py via command line uses your local python environment. I wonder if package compatibility has to do with some of the issues. Which python version are you using? I can send along a windows standalone compiler conda environment to you.

duserzym:
Ah I remember the .remove() issue now. I saw this issue after upgrading matplotlib 7 months ago. I fixed this when compiling standalones but never committed the change to the main repo. If we change this, people using older matplotlib versions will also have issues when opening up demag gui via command line. I didn't know what to do. I don't like package updates like this!

Swanson-Hysell:
Thanks for fixing these issues @apivarunas. I think it makes sense to merge it in nowso that we have a version working with the current install instructions.

apivarunas:
@Swanson-Hysell If issues do pop up in the aftermath, tag me and I'll be happy to try and fix them!

Issue #668, opened 2022-10-26
ipmag.plot_di having line between datapoints


sslotznick:
In PmayPy Gui, you are able to plot equal area plots for a single specimen with a line that connects that data points in order of their measurement-demagnetization sequence.  (See example below.)

Would it be possible to add this line plotting option to the PmagPy ipmag.plot_di equal area?

[H16-1A.1_EA_tilt.pdf](https://github.com/PmagPy/PmagPy/files/9872242/H16-1A.1_EA_tilt.pdf)

Swanson-Hysell:
Certainly possible and would be good to implement. A goal of ours is to have more interoperability between the Demag_GUI functions and a the ipmag ones — this is a good illustration of why that will be useful.

duserzym:
This output function in `demag_gui.py `is wrapped in its own function called `draw_interpretations`. 

You might want to check `ipmag.zeq_magic` which makes zijderveld and equal area plots for magic formatted measurements files. It can plot interpretations. I haven't used this function. Perhaps we can play with the specimen file in the input to plot only selected specimens.

apivarunas:
Reviving a slightly necrotic issue, because I was going through looking for things to help with. I also thought this might be nice to have, and I think I've solved it.

For the ipmag.plot_di command are you (@sslotznick) looking for something like so:

![image](https://github.com/user-attachments/assets/fb1921dd-3543-49a6-9ce7-db699815ed1f)
connected

![image](https://github.com/user-attachments/assets/ab3a6fc4-e334-46d8-99d3-a419f4c39b15)
unconnected

If so, I can push a commit adding the connection as an option to ipmag.plot_di pretty soon.

sslotznick:
@apivarunas yes that is what I was looking for!  

Line could be a touch lighter in color or thickness but that is easy enough to modify post!

apivarunas:
Resolved with #724.

Swanson-Hysell:
Thanks for resolving this issue @apivarunas!

PR #724, opened 2025-01-11
Resolving issue #668 connect points

apivarunas:
This pull request fixes issue #668. Added capability to ipmag.plot_di to be able to connect points by using a "connect_points" flag in the function call. Flag is set to False, so the capability is an opt-in. Works for normal directions and reverse directions but does not connect normal directions _to_ reverse directions. I think that's fine.

Linewidth is 0.25 pt, and color is black for the most neutrality. Linewidth was thinned after feedback.

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -2211,12 +2211,12 @@ def plot_di(dec=None, inc=None, di_block=None, color='k', marker='o', markersize
 
     if len(X_up) > 0:
         if connect_points == True:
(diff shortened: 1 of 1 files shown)


duserzym:
How about we also add optional arguments to control connecting line color, width, and alpha?

apivarunas:
Yeah, that works. I contemplated it but thought that the function was getting a little busy with arguments. But I guess they're mostly silent defaults.

Issue #705, opened 2023-09-27
ipmag.eqarea_magic not saving plots to specified dir_path


Swanson-Hysell:
**What command was run:**

ipmag.eqarea_magic(save_plots=True, 
                   input_dir_path=directory,
                   dir_path=directory,
                   )

**What happened:**

Plots were saved to local directory rather than specified directory in dir_path.

**What I expected to happen:**

Plots to be saved to specified directory.

**Data file(s):**

Can reproduce within Inspecting_MagIC_directional_data.ipynb pmagpy-docs notebook

**Screenshot(s):**

<img width="583" alt="Screenshot 2023-09-27 at 11 09 06 AM" src="https://github.com/PmagPy/PmagPy/assets/4332322/282769f4-7bb4-4556-99b1-f35fc755191f">

**PmagPy version:**

pmagpy-4.2.115

As an additional point, it would be nice to add a dir_path option to ipmag.zeq_magic() which does not have one.

apivarunas:
I was interested in this, so I tracked down the dir_path issues. Should be fixed now with #725. On a side note, how does one edit the docs? I have some fixes for the relevant notebook.

Swanson-Hysell:
Nice work with the fix @apivarunas!

PR #725, opened 2025-02-02
Adding ability to save plots to specified directory

apivarunas:
Dealing with issue #705. I tracked this down to pmagplotlib.py, the function `pmagplotlib.save_plots`. No directory was specified in that function so it would just save to local. I have now added the ability for both `ipmag.eq_area_magic` and `ipmag.zeq_magic` to send the dir_path to pmagplotlib and the files get saved to the correct directory.

I tested it in the Inspecting_MagIC_directional_data.ipynb pmagpy-docs notebook*, and it worked.

*The Inspecting_MagIC_directional_data.ipynb pmagpy-docs notebook does not run very well in its present state. The function `ipmag.download_magic_from_doi()` does not have the correct parameters, and I don't think there is a function `ipmag.list_MagIC_names()`. I had to change the initial function to `ipmag.download_magic_from_id()` which seemed to better match the relevant inputs and outputs. The MagIC ID in question is "19640".

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -12346,7 +12346,7 @@ def zeq_magic(meas_file='measurements.txt', spec_file='',crd='s', dir_path = "."
     dir_path : str
         output directory for plots, default "."
     input_dir_path : str
-        input directory, if different from dir_path, default "."
+        input directory, if different from dir_path, default ""
     angle : float
         angle of X direction with respect to specimen X
     n_plots : int, default 5
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -12327,7 +12327,7 @@ def atrm_magic(meas_file, dir_path=".", input_dir_path="",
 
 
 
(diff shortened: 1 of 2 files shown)


Swanson-Hysell:
Thanks for working on this.

Why does the code have the parameter in the function be `input_dir_path=""` rather than `input_dir_path="."` that is stated as the default?

apivarunas:
Ah, good catch. It's because I was imitating the `ipmag.eqarea_magic()` parameters but then made a typo in the `ipmag.zeq_magic()` docstring. They should match, now. Still functional.

apivarunas:
Anything else you want done for this issue @Swanson-Hysell ? I can play around with the docs a bit, but I figure I'll leave that for a separate commit.

Swanson-Hysell:
Thanks for this! Just tested it out and it works well. I made one little tweak to make sure that it also works when a dpi is specified.

Swanson-Hysell:
I saw in the issue that you found the PmagPy-docs repo — committing an updated notebook that fixes issues there with a pull request would be great. Once a commit is made, the online JupyterBook docs auto regenerate.

Issue #709, opened 2024-02-05
implement midpoint derivative calculation


Swanson-Hysell:
Currently within `ipmag.curie()` there is code to calculate the derivative which is then smoothed. It seems preferable to use the midpoint temperature (i.e. `T_mid = (T[i + 1] + T[i]) / 2`) for the derivative rather than the original T array values in order to associate the derivative value between the two points from which it is calculated.

Here is the current relevant code:
```
    # calculate first derivative
    d1, T_d1 = [], []
    for i in range(len(M_smooth) - 1):
        Dy = M_smooth[i - 1] - M_smooth[i + 1]
        Dx = T[i - 1] - T[i + 1]
        d1.append(Dy/Dx)
    T_d1 = T[1:len(T - 1)]
    d1 = np.array(d1, 'f')
    d1_smooth = smooth(d1, window_len)
```

apivarunas:
@Swanson-Hysell I took a look at this issue. I've actually never used the `ipmag.curie()` function since the function initially enforces even time-steps (=1.0) or quits. And I don't know about other Kappabridge machines, but I know that neither I've used has done that. I believe the function is able to handle and interpolate irregular data, one just needs to reorder the initial statements. I've almost got it working with a test file, I just have one last bug that must be related to the length of the array since it only crops up on the last pair of values.

I'll send out a PR related to that soon-ish.

**Digging into the actual raised issue**, I think that the original T array values do in fact represent the midpoint, since the calculation of the derivative is going from `Dx = T[i - 1] - T[i + 1]`. That can be adjusted to be consecutive M/T steps and the midpoint  implemented. Would you like that to be the case?

Swanson-Hysell:
Good point here. The current approach is calculating Dx across the step before and after the temperature. This approach actually implements a bit of smoothing by spanning more temperatures steps rather than just consecutive ones. Given that there is already a pre-smoothing step, it seems like it could make sense to modify the function to be consecutive M/T steps with the midpoint implemented. Does that approach make the most sense to you?

PR #726, opened 2025-02-15
implementing #709 requests

apivarunas:
So, I made a few changes.
1. I implemented the derivative calculation using np.gradient() [docs](https://numpy.org/doc/stable/reference/generated/numpy.gradient.html). This uses second-order accurate central differences for interior points. It also runs over the entire array which removes some of the messy T-1, T-2 parametesr throughout. Running the test case: `ipmag.curie(file_name='curie_example.dat', window_length=10)` shows identical behavior to the previous implementation with a slight difference in the maximum probably related to the T-array shift.

2. I switched order of two docstring parameters to mirror the function and put 10 as the default window length.

3. I commented out the `sys.quit()` line (so it's now just a warning, and doesn't quit the function) and moved that warning farther down the chain of loops to allow the earlier data interpolations to work.

4. Two minor changes to account for numpy updates (np.float64 in `ipmag.curie()` and np.ones in `ipmag.smooth()`.

I've also tested this with an irregularly spaced example file [curie_irregular.txt](https://github.com/user-attachments/files/18812349/curie_irregular.txt), and it works as well. Kicks out one warning for the last pair of values, but behaves otherwise.

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -8105,7 +8105,7 @@ def smooth(x, window_len, window='bartlett'):
 
     # s=numpy.r_[2*x[0]-x[window_len:1:-1],x,2*x[-1]-x[-1:-window_len:-1]]
     if window == 'flat':  # moving average
-        w = ones(window_len, 'd')
+        w = np.ones(window_len, 'd')
     else:
         w = eval('np.' + window + '(window_len)')
     y = np.convolve(w/w.sum(), s, mode='same')
@@ -8113,7 +8113,7 @@ def smooth(x, window_len, window='bartlett'):
 
 
 def curie(path_to_file='.', file_name='', magic=False,
-          window_length=3, save=False, save_folder='.', fmt='svg', t_begin="", t_end=""):
+          window_length=10, save=False, save_folder='.', fmt='svg', t_begin="", t_end=""):
     """
     Plots and interprets curie temperature data.
     The 1st derivative is calculated from smoothed M-T curve (convolution
@@ -8124,9 +8124,9 @@ def curie(path_to_file='.', file_name='', magic=False,
     Temperature steps should be in multiples of 1.0 degrees.
 
     Parameters:
-        file_name : name of file to be opened
         path_to_file : path to directory that contains file (default is current directory, '.')
-        window_length : dimension of smoothing window (input to smooth() function)
+        file_name : name of file to be opened
+        window_length : dimension of smoothing window (input to smooth() function, default=10)
         save : boolean argument to save plots (default is False)
         save_folder : relative directory where plots will be saved (default is current directory, '.')
         fmt : format of saved figures (default is svg)
@@ -8148,7 +8148,7 @@ def curie(path_to_file='.', file_name='', magic=False,
         magn_key = cb.get_intensity_col(data_df)
         M = data_df[magn_key].values
     else:
-        Data = np.loadtxt(complete_path, dtype=np.float)
+        Data = np.loadtxt(complete_path, dtype=np.float64)
         T = Data.transpose()[0]
         M = Data.transpose()[1]
     T = list(T)
@@ -8170,10 +8170,6 @@ def curie(path_to_file='.', file_name='', magic=False,
     # exit if deltaT is not integer
     i = 0
     while i < (len(T) - 1):
-        if (T[i + 1] - T[i]) % 1 > 0.001:
-            print("delta T should be integer, this program will not work!")
-            print("temperature range:", T[i], T[i + 1])
-            sys.exit()
         if (T[i + 1] - T[i]) == 0.:
             M[i] = np.average([M[i], M[i + 1]])
             M.pop(i + 1)
@@ -8188,6 +8184,10 @@ def curie(path_to_file='.', file_name='', magic=False,
                 M.insert(i + 1, slope * (T[i] + 1.) + b)
                 T.insert(i + 1, (T[i] + 1.))
                 i = i + 1
+        elif (T[i + 1] - T[i]) % 1 > 0.001:
+            print("delta T should be integer, be careful!")
+            print("temperature range:", T[i], T[i + 1])
+            #sys.exit()
         i = i + 1
 
     # calculate the smoothed signal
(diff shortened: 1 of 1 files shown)


apivarunas:
@Swanson-Hysell If this looks okay, would you like this implemented in `rockmagpy` or are you pretty happy with that implementation?

Swanson-Hysell:
Can you help me understand the warning saying that *delta T should be integer, be careful!* given that the delta T is irregular throughout the data set?

There is a broader discussion here about implementing varied approaches in the context of this function:
- peak of first derivative
- two tangent method
- second derivative zero crossing
Can you think about adding these different approaches and having there be parameters that can do one method or another?

The goal is to avoid duplicative functions between `rockmagpy` and other `pmagpy` libraries is the `ipmag` implementation is what we can be using in `rockmagpy`.

Swanson-Hysell:
There appears to be a fundamental issue in this ipmag.curie() function given that it is searching for the peak of the 2nd derivative. One could search for the peak of the first derivative, but this isn't appropriate for the second derivative which instead should be the zero crossing (as in this Verwey temperature example: https://pmagpy.github.io/RockmagPy-notebooks/MPMS_verwey_fit.html). In this example (and the other example file) the data are too noisy for this approach (two tangent is likely better).

There is another method to implement which is the linearity of inverse susceptibility when dealing with high-temperature susceptibility data:
Screenshot 2025-02-21 at 2 14 17 PM

https://doi.org/10.1029/2006JB004507

apivarunas:
> Can you help me understand the warning saying that _delta T should be integer, be careful!_ given that the delta T is irregular throughout the data set?

Yeah. Well, sort of. I believe the input arrays are being regularized (delta = 1) up until the last pair of values. 

I don't think that this has much affect on the behavior of the function result, so we could remove the warning, I suppose. Better yet would be to figure out how to continue the regularization through the whole array...which I'm working on.

> There is a broader discussion here about implementing varied approaches in the context of this function:
> 
> * peak of first derivative
> * two tangent method
> * second derivative zero crossing
>   Can you think about adding these different approaches and having there be parameters that can do one method or another?

Yes. I think that's a good idea. Including your most recent comment as far as approaches goes. I can work on this over the weekend.

> The goal is to avoid duplicative functions between `rockmagpy` and other `pmagpy` libraries is the `ipmag` implementation is what we can be using in `rockmagpy`.

Right. I think eventually the function should have some comprehensive approaches and only needs to live in one module.

apivarunas:
> There appears to be a fundamental issue in this ipmag.curie() function given that it is searching for the peak of the 2nd derivative. One could search for the peak of the first derivative, but this isn't appropriate for the second derivative which instead should be the zero crossing (as in this Verwey temperature example: https://pmagpy.github.io/RockmagPy-notebooks/MPMS_verwey_fit.html). In this example (and the other example file) the data are too noisy for this approach (two tangent is likely better).

Right. That's been the implementation in the software for a while. I think, looking at [Fabian et al., 2013]( https://doi.org/10.1029/2012GC004440) Figure 8 that it probably does okay. But I think adjusting it is in our best interest going forward. 

So, a summary on the methods to implement:

1. I think we'd want the minimum of the first derivative ([Fabian et al., 2013]( https://doi.org/10.1029/2012GC004440))

image

2. I was pretty convinced by ([Petrovsky and A. Kapicka, 2006](https://doi.org/10.1029/2006JB004507)) that the two tangent method is _not_ the correct way to go.

3. Zero crossing is fine, but problematic when the data is noisy. Honestly, it seems as if the smoothing function isn't doing too much until a pretty impressive large window size is reached.

It would be pretty easy to modify the function to address 1 and 3 together, really just a matter of reporting the minimum of the first derivative instead of the maximum of the second derivative. I have to see what exactly is up with the smoothing function.

> There is another method to implement which is the linearity of inverse susceptibility when dealing with high-temperature susceptibility data: Screenshot 2025-02-21 at 2 14 17 PM
> 
> https://doi.org/10.1029/2006JB004507

Yes, this can be added as well, I think. I'll play around with these additions.

apivarunas:
@Swanson-Hysell

Okay, whew. I've rewritten the function so that it gives the Curie point estimate based on first derivative min and second derivative zero crossing.

Also, the interpolation for irregular input is better, that error is gone.

The plotting is now done by `matplotlib`, since the `plot_xy()` function in `pmagplotlib` is, according to the docstring, deprecated. 

I don't think the two tangent method is necessary, since it isn't particularly good for susceptibility data. The current iteration is probably a viable point to provide an update to the ipmag.curie() function. I haven't implemented the inverse susceptibility operability yet. Trying to figure out if it makes sense to let a user pick a linear section or to linefind through the 1/kT curve for them. Options, options.

apivarunas:
@Swanson-Hysell I have now added the option to use inverse susceptibility. It is called by setting `inverse_sus = True` in the function (default is False).

There is an additional import added to `ipmag.py`:

`from scipy.stats import linregress`

And a small additional utility function called `ez_line` which uses `matplotlib` to plot a line from slope and intercept on the graph.

The basic flow of the addition is:

1. Plot inverse susceptibility.
2. Ask user to provide the high temperature window they want to examine (since its hard to view the whole curve).
3. Plot that window.
4. Ask user to provide the high and low temperature values that define the linear segment.
5. Regress a line through that segment, calculate its parameters. Return the x-intercept and R-squared value, and plot the window, values used in the regression, and line (image of final output shown below).

image

Swanson-Hysell:
I will dig into this tomorrow. What are your thoughts on having this as an option within a single function (which I know is what I originally suggested) vs having it be its own function. Given how many specific parameters for the inverse susceptibility analysis the user needs to provide, it seems like it might work better was its own function.

apivarunas:
The overall function does end up pretty busy. However, there isn't much the user needs to _do_, necessarily. The _inverse_sus_ flag is probably the most involved piece by a fair degree.

So...yeah. I think its reasonable break it out into its own function. Let me know if you want that to happen and it should be easy enough to implement.

apivarunas:
Okay, the inverse method is now broken out into its own function as suggested.

apivarunas:
Any improvements wished, @Swanson-Hysell?

apivarunas:
This PR request is out of date due to the incorporation of the `rockmag` module. As discussed in #802, curie should probably move out of `ipmag`. #776 incorporates the useful part of this PR.

PR #734, opened 2025-04-08
Fixing circular calls in two functions

apivarunas:
Calling `ipmag.fishrot()` several times inside ipmag.py doesn't work very well! Fixed. Also, added a few extra spaces to comments, and removed a redundant parentheses.


Swanson-Hysell (approved):
Nice catch. Good to fix this!

Issue #673, opened 2022-11-14
edit docstrings in Google format for documentation rendering


Swanson-Hysell:
We have found that our modified Numpy-style docstrings are not rendered particularly well within the automobile API documentation generation that we are running for our documentation.

As a result, @duserzym and I experimented and found that Google-style Python docstrings (https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) render well. An advantage of these docstrings are that they are also more easily read within the code as well (then rigid Numpy docstrings which is why we were doing some hybrid Numpy docstrings that weren't getting rendered well).

Given the importance of rendering the docstrings within the new documentation (https://pmagpy.github.io/PmagPy-docs/api/api.html), it makes sense to progressively make edits to translate our docstrings into this Google-style format.

A doc string that looks like this:

```
def fisher_mean(dec=None, inc=None, di_block=None):
    """
    Calculates the Fisher mean and associated parameters from either a list of
    declination values and a separate list of inclination values or from a
    di_block (a nested list of [dec,inc,1.0]). Returns a dictionary with the 
    Fisher mean and statistical parameters.

    Parameters:
        dec : list of declinations or longitudes
        inc : list of inclinations or latitudes
            or
        di_block : a nested list of [dec,inc,1.0]
            A di_block can be provided instead of dec, inc lists in which case it
            will be used. Either dec, inc lists or a di_block need to be provided.

    Returns:
        dictionary 
            Dictionary containing the Fisher mean parameters.
            This dictionary can be printed in a more readable fashion using the
            ``ipmag.print_direction_mean()`` function if it is a directional mean or 
            ``ipmag.print_pole_mean()`` function if it is a pole mean.

    Examples:
        Use lists of declination and inclination to calculate a Fisher mean:

        >>> ipmag.fisher_mean(dec=[140,127,142,136],inc=[21,23,19,22])
        {'alpha95': 7.292891411309177,
        'csd': 6.4097743211340896,
        'dec': 136.30838974272072,
        'inc': 21.347784026899987,
        'k': 159.69251473636305,
        'n': 4,
        'r': 3.9812138971889026}

        Use a di_block to calculate a Fisher mean (will give the same output as the
        example with the lists):

        >>> ipmag.fisher_mean(di_block=[[140,21],[127,23],[142,19],[136,22]])
    """
    if di_block is None:
        di_block = make_di_block(dec, inc)
        return pmag.fisher_mean(di_block)
    else:
[...]

PR #735, opened 2025-04-17
Testing AFP new setup of VS Code

apivarunas:
Simplifying some rather flowery docstrings written by myself in order to test out my new installation of VS Code from editing, to commiting, to pushing, to pulling. I think I like it better than PyCharm.


apivarunas:
Tested the full process from edit to merging to master on the rather safe pursuit of editing docstrings as in #673. Let me know if this is how the process should go, @Swanson-Hysell. I think I'm safe merging docstrings, but for more complicated stuff I'll take it easy for a while.

Swanson-Hysell:
Looks good. Glad you are enjoying the VSCode lifestyle.

Issue #736, opened 2025-04-18
SyntaxWarning: invalid escape sequence upon import in Python 3.12


Swanson-Hysell:
**What command was run:**

Import PmagPy

**What happened:**

Got numerous SyntaxWarnings

```
[/Users/unimos/0000_Github/PmagPy/pmagpy/contribution_builder.py:1467](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/pmagpy/contribution_builder.py:1467): SyntaxWarning: invalid escape sequence '\d'
  """
[/Users/unimos/0000_Github/PmagPy/pmagpy/contribution_builder.py:1498](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/pmagpy/contribution_builder.py:1498): SyntaxWarning: invalid escape sequence '\d'
  prog = re.compile("\d*[.]\d*([0]{5,100}|[9]{5,100})\d*\Z")
[/Users/unimos/0000_Github/PmagPy/pmagpy/contribution_builder.py:2033](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/pmagpy/contribution_builder.py:2033): SyntaxWarning: invalid escape sequence '\s'
  pattern = re.compile('{}(?=:|\s|\Z)'.format(meth_code))
[/Users/unimos/0000_Github/PmagPy/pmagpy/ipmag.py:6813](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/pmagpy/ipmag.py:6813): SyntaxWarning: invalid escape sequence '\d'
  pattern = re.compile('[4][-]\d')
[/Users/unimos/0000_Github/PmagPy/pmagpy/ipmag.py:6820](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/pmagpy/ipmag.py:6820): SyntaxWarning: invalid escape sequence '\d'
  pattern = re.compile('[7][-]\d')
[/Users/unimos/0000_Github/PmagPy/pmagpy/ipmag.py:8470](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/pmagpy/ipmag.py:8470): SyntaxWarning: invalid escape sequence '\h'
  sf.write('\hline\n')
[...]

PR #740, opened 2025-04-24
Raw-ified strings in file

apivarunas:
The remainder of #736, fixing strings in `contribution_builder.py `. I'm not exactly sure why line 1467 is throwing a warning, considering that it's the docstring quote. Apart from that, removed a few lines of dead space in the file.


apivarunas:
And I remember something that I may have been told already some other evening...docstrings are string-y so they can get the same SyntaxError.

PR #741, opened 2025-04-24
svei.py compatibility, duplicate functions

apivarunas:
Okay, so, I decided to use the `svei.py` example notebook and ran into a few things in `svei.py`. I think I've made the correct adjustments. 

Descriptions:

1. The `scipy.integrate` import of `cumtrapz` did not work. This was due to a name change in `scipy` release 1.6.0 to `cumulative_trapezoid`. I adjusted this import and the subsequent usage in the file.

2. Errors cropped of from use of `np.int()`, which is deprecated in favor of the built-in `int` as of `numpy` 1.20.0. I adjusted these to `int`

3. While renaming the instances, I saw that the function `svei.GGP_cdf()` was duplicated. I removed the duplicate.

4. I was curious about other duplicates and found that `svei.AD_dec` and `svei.AD_inc` were also duplicated. Since the implementations of these were identical except for more complete docstrings on one of each pair, I removed the identical functions which lacked docstrings.

### DIFF
--- a/pmagpy/svei.py
+++ b/pmagpy/svei.py
@@ -6,7 +6,7 @@ import matplotlib.pyplot as plt
 import pmagpy.pmag as pmag
 import pmagpy.ipmag as ipmag
 from scipy.interpolate import PchipInterpolator
-from scipy.integrate import cumtrapz
+from scipy.integrate import cumulative_trapezoid
 from scipy import interpolate
 from scipy.stats.distributions import chi2
 import pandas as pd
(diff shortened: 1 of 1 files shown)


ltauxe:
Thanks Tony! 
lisa

Swanson-Hysell:
Would it be ok if we moved SVEI_demo.ipynb into its own repository within the PmagPy organization? Seeking to run the notebook in the main repo PmagPy leads to issues with the directories:
Screenshot 2025-04-30 at 1 28 23 PM
For this code, a folder named 'PSV10-24/' needs to be created within the folder where the notebook lives for example.

apivarunas:
I suppose we could adjust the code in the demo notebook to fix that directory issue. However, moving some of the example notebooks to their own repository would be "clean", I guess. I bet there are some complexities to it I'm not immediately aware of though.

Issue #759, opened 2025-06-12
y-axis wrong on unmixed coercivity plots


sslotznick:
**What command was run:**
rmag.interactive_backfield_fit(MaxUnmix_experiment['smoothed_log_dc_field'], 
                               MaxUnmix_experiment['smoothed_magn_mass_shift'], 
                               n_components=2, figsize=(10, 4), skewed=False,)

**What happened:**
A plot was created with a y-axis of dM/dB

**What I expected to happen:**
In fact since it is a log scale it should be dM/dlog(B)

apivarunas:
As all instances of this have been fixed by @sslotznick, I think we're good here. Thanks!

PR #760, opened 2025-06-12
Update interactive_backfield_fit to have y-axis in plot right

sslotznick:
Added dM/dlogB instead of dM/dB


sslotznick:
This partially solves #759.

PR #768, opened 2025-06-20
Add find_svei_kent function

anthony-j-fuentes:
This function is an extension of the find e/I Kent function that uses the f-factor bounds output by the svei method of  Tauxe et al., (2024).


Swanson-Hysell:
@duserzym Here is the comparison of the 95% contour (in black) with the Kent mean in blue for the Cutface data. Pretty good agreement:  Screenshot 2025-06-20 at 9 51 42 AM

Issue #758, opened 2025-06-12
return axis for every plot function


outfrenk:
**Is your feature request related to a problem? Please describe.**
Some plots do not return an axis, making it difficult to change the plot to my liking afterwards

**Describe the solution you'd like**
Always return the axis from the plot

**Additional context**
It could also be an option incorperated in the plot functions

apivarunas:
@outfrenk Thanks for opening this issue. I'm happy to work on this. Could you provide an initial example of a function that does what you want, and a function that doesn't do what you want?

Are you basically looking for a save_plots call in all functions that produce a plot? I guess some functions produce plots, but do not save them?

Swanson-Hysell:
This issue was raised with respect to functions in rockmag.py which is a good place to start with addressing this issue. 

The idea is that we want to have the option to return the fig and axis objects associated with each plotting function so that the user can further customize the plots (i.e. adjust axis limits).

There is a discussion to be had about defaults. It could make sense to have the default be to not return the the figure with the idea that it is being displayed inline in the notebook with advanced users knowing to put in `return_figure=True` as a parameter. That is how it is currently implemented in 
`ms_t_plot`:

```
def ms_t_plot(
    data,
    temperature_column="meas_temp",
    magnetization_column="magn_mass",
    input_unit="K",
    plot_unit="K",
    interactive=False,
    return_figure=False,
    show_plot=True,
    size=(6, 3),
    legend_location="upper left",
):
    """
    Plot magnetization versus temperature in static or interactive mode.

    Parameters
    ----------
    data : pandas.DataFrame or array-like
        Table or array containing temperature and magnetization data.
    temperature_column : str, default 'meas_temp'
        Name of the temperature column in `data`.
    magnetization_column : str, default 'magn_mass'
        Name of the magnetization column in `data`.
    input_unit : {'K', 'C'}, default 'K'
        Unit of the input temperature data.
    plot_unit : {'K', 'C'}, default 'K'
        Unit for the x-axis display.
    interactive : bool, default False
        If True, use Bokeh for an interactive plot.
    return_figure : bool, default False
        If True, return the figure object(s). Assign to capture, e.g.:
            fig, ax = ms_t_plot(..., return_figure=True)
    show_plot : bool, default True
        If True, display the plot immediately.
    size : tuple(float, float), default (6, 3)
        Figure size in inches (Matplotlib) or height for Bokeh.
    legend_location : str, default 'upper left'
        Legend location in Matplotlib terms.

[...]

PR #770, opened 2025-08-08
Initial work on #758

apivarunas:
Okay, so I've started working on #758. I've done a few things here:

1. Changed two function names. To wit: `day_plot` changed to `plot_day` and `day_plot_MagIC` changed to `plot_day_MagIC`. **But why?** So that all plotting functions have a similar name, plot_etc

2. Had `plot_day` now create both a fig, ax and return them. This mirrors the implementation in `ms_t_plot` (and you know I want to change that function name as well...) as discussed in #758 . The default return on the fig, ax, however, is `True` so that they can flow seamlessly into `plot_day_MagIC`.

Apart from that, adding to the docstrings for `plot_day`, etc.

### DIFF
--- a/pmagpy/rockmag.py
+++ b/pmagpy/rockmag.py
@@ -4555,9 +4555,9 @@ def add_Bcr_to_specimens_table(specimens_df, experiment_name, Bcr):
     return     
 
 
-# Day plot function
+# Day plot functions
 # ------------------------------------------------------------------------------------------------------------------
-def day_plot_MagIC(specimen_data, 
+def plot_day_MagIC(specimen_data, 
                    by ='specimen',
                    Mr = 'hyst_mr_mass',
                    Ms = 'hyst_ms_mass',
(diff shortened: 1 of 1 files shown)


apivarunas:
Okay, now have additionally renamed neel_plot to plot_neel(), and neel_plot_magic to plot_neel_magic for standardization. Additionally, plot_hysteresis loop now has an option to return static (Matplotlib) axes, as well as returning axes in any case.

apivarunas:
@Swanson-Hysell Okay, I think this incorporates axes return for a few functions, and does some re-naming of plot functions to make them more consistent. All plot functions which return axes now start `plot_`. A few interactive plotting functions now start with `interactive_` and do not return axes.

If the goal is to get all those interactives to return axes as well, then I can do that, just let me know. I figure this is a good spot to take stock, get a review, and merge in this PR, however.

Swanson-Hysell:
`interactive_MPMS_signal_blender` could just be `mpms_signal_blender`

apivarunas:
I've renamed as requested, I also renamed one other function with a similar setup (i.e., `goethite_removal_interactive` which has an identical leading name as `goethite_removal`. So in both cases the autocomplete should be our friend.

Swanson-Hysell (approved):

PR #775, opened 2025-08-30
utf-8 Encoding issue

apivarunas:
Trying to run a notebook (`high_T_susceptibility.ipynb`) which was importing data from the MagIC database, I was occasionally running into the error:

> UnicodeEncodeError: 'charmap' codec can't encode character '\ufffd' in position 236: character maps to <undefined>

Looking up the error, it seems that it's an issue primarily on Windows, which doesn't use utf-8 by default. By adding the flag `encoding="utf-8"` to line 5246, the problem was fixed, and I was able to import the data without a problem. This problem likely doesn't show up on Macs. For good measure, I fixed other lines with a similar `with open()` structure.

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -3551,7 +3551,7 @@ def combine_magic(filenames, outfile='measurements.txt', data_model=3, magic_tab
                 magic_table))
             return False
         # figure out file type from first of files to join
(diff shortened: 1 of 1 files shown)


Swanson-Hysell:
Thanks for checking and fixing!

PR #780, opened 2025-10-05
Removing functions deprecated in 2022

apivarunas:
As I was perusing the docstrings in `pmagpy.pmag`, I saw a number of functions that were marked as "depreciated" as of 2022. These are:

`vspec`
`upload_read`
`scoreit`
`dread`
`Dir_anis_corr`
`int_pars`
`mark_demag_rec`
`apseudo`
`doaniscorr`
`dolnp3_0` (instead, dolnp is recommended)
`getnames` (commented out)

I imagined these weren't load bearing. However, on a next-day review before this PR, I found that two functions, `int_pars` and `apseudo`, are still used inside other functions. So, I added them back in. The others aren't used elsewhere.


Swanson-Hysell:
So I think that we should tread carefully here and look through each function.

`dolnp3_0` seems like a clear case for removal given Lisa's note 7 years ago of:
>     DEPRECATED!!  USE dolnp()

and that it isn't used elsewhere in the codebase.

apivarunas:
> I don't know why @tcarrasc marked `pmag.apseudo()` as deprecated. It is central to how `pmag.s_boot()` is working which is then central to AMS bootstrapping throughout the project. It seems important to remove that deprecation notice (and I am curious why it was added).
> 
> Similar for `pmag.int_pars` which is central to creating the dictionary that `pmag.PintPars()` builds upon.
> 
> I think that @tcarrasc was focused on making docstrings and may have overlooked these uses of the functions. It could make sense to remove the deprecation notice and at the same time create new associated docstrings.

Good point. I added those two back in since they were clearly used in other functions. I'll work on docstrings and removing the deprecation notice for both of them.

But right, it does mean a more robust check on all of them is warranted, I guess.

apivarunas:
I've given this some time to sit and checked back through to make sure that functions used elsewhere were not removed. At the very least, the elsewhere-used functions now aren't marked as deprecated...

apivarunas:
Unfortunately, this has created some problems. `dolnp3_0` _was_ actually used, in `demag_gui.py`. Twice, in fact. Don't know how I didn't see that. However, `dolnp3_0` still ends up using `dolnp`, it just does some different data pre-processing. And `dolnp` has a different argument structure. 

Lovely...

Basically, causes problems with trying to save Magic tables.

Swanson-Hysell:
It seems like we can simply add that function back in for now. This pull request does it: https://github.com/PmagPy/PmagPy/pull/860 Can you have a look at that @apivarunas and make sure I am doing the right thing? And then merge in if it looks good.

apivarunas:
That looks like what I did to fix it locally, which seems to work. However, this made me notice something different.

I had a few sites of young volcanic rocks that didn't demagnetize effectively with AF. So, I ran thermal demagnetization on them. That's actually what I've using the GUIs for right now. The data still ended up being single-component in nearly all cases, so I made fits which used both the AF and thermal demagnetization data. However, when I save the MagIC Tables, the program seems to delete(!) the fits which straddle both AF and thermal demagnetization. As in, when I re-open the GUI, the only fits which are preserved are those which only have AF or only have thermal steps.

The 'sites' table, then, preserves calculated means which only take into account those sole-demagnetization method fits.

Swanson-Hysell:
@apivarunas That behavior certainly deserves its own issue! I think that this could be associated with how bounds are currently defined and used within the code. That is that you define a fit by the lower bound of the demag step and the upper bound of the demag step. When you save a redo file and then import it, do you have the same issue?

apivarunas:
Actually, I've dug into it more and it specifically occurs when reopening a contribution where the multi-method fits have been saved as Magic tables. I've also tracked down a comment referencing it, I think.

Additionally, I've tracked down a little bug in averaging a site mean by samples, and managed to safely eliminate dolnp3_0 after all. I'm going to make a PR this evening once I've done a little more testing.

Then dig into this other reopening thing.

Issue #794, opened 2025-10-08
error in generating title in ipmag.find_ei()


Swanson-Hysell:
**What command was run:**

```
data=np.loadtxt('data_files/find_EI/find_EI_example.dat')
ipmag.find_ei(data)
```

**What happened:**

This error resulted which I have bumped into a couple of times:

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
/var/folders/57/lmgz7h4x11v34k53c6ldrlrw0000gn/T/ipykernel_10973/1651072254.py in <module>
      1 data=np.loadtxt('data_files/find_EI/find_EI_example.dat')
----> 2 ipmag.find_ei(data)

~/0000_GitHub/PmagPy/pmagpy/ipmag.py in find_ei(data, nb, save, save_folder, fmt, site_correction, return_new_dirs, figprefix, return_values, num_resample_to_plot, data_color, EI_color, resample_EI_color, resample_EI_alpha, tight_axes)
   9415         title = 'Pathological Distribution: '
   9416     else:
-> 9417         title = '%7.1f [%7.1f, %7.1f]' % (Inc, I[lower], I[upper])
   9418     if save:
   9419         plt.savefig(save_folder+'/'+figprefix+'_EI_bootstraps'+'.'+fmt, bbox_inches='tight', dpi=300)

IndexError: list index out of range
```

The analysis looks to be moving forward ok, but something is going on with the generation of the title for the plot.

**Screenshot(s):**

<img width="655" height="621" alt="Image" src="https://github.com/user-attachments/assets/6368f625-78fe-44d9-9bf2-ce0439eea05b" />

**PmagPy version:**

4.3.7

PR #795, opened 2025-10-08
change from predetermined index for percentiles to direct percentile calculation in find_ei and find_ei_kent

Swanson-Hysell:
The goal here is to fix the bug that is documented in:
https://github.com/PmagPy/PmagPy/issues/794
where the index is not found associated with find_ei inclinations. Rather than using the predetermined index, the percentile is calculated from the inclinations given that the total of successful bootstraps can be different than the specified number.

This fix appears to fix the bug within the `PmagPy_plots_analysis.ipynb` `ipmag. find_ei()` example. Can you please review and test this fix within this branch @duserzym?

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -9345,10 +9345,10 @@ def find_ei(data, nb=1000, save=False, save_folder='.', fmt='svg',
 
     Returns:
         - equal area plot of original directions
-        - Elongation/inclination pairs as a function of f,  data plus 25 bootstrap samples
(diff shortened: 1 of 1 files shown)


apivarunas:
@Swanson-Hysell I checked on this PR and the fix works correctly. The code works correctly now. I think it's good to go with a merge.

Aside: One probably 'nothing' thing I noticed, though, was the numbers on one of the generated plots:

image

Now, I understand that the title is the corrected inclination [low_bound     high_bound]. However, given the plot, it looks as if the numbers are labeling the lines, and thus that the corrected inclination and low bound are swapped. I know that's not the case, but it caught my eye.

Issue #480, opened 2018-12-06
forc_diagram has lots of warnings on Windows


moonshoes87:
but the plot seems to be generated ok.

apivarunas:
#811 now fixes this issue by pre-emptively catching mismatched shapes or NaN values in the internal `d2_func`. The code now runs without throwing those errors. This can be closed once that is merged.

apivarunas:
All done, via #811. The fix also includes a warning that such a thing is occurring.

PR #811, opened 2025-11-10
Fixing issues with Jupyter FORC plotting

apivarunas:
When running the example file, I had some issues running the `forc_diagram` example. I believe this was due to some issues with the %matplotlib inline command in the program and the TKAgg backend being set in the program. I fixed this in the imports, where the program now checks where it is being called from and sets a backend based on this.

As I dove into this, there was some redundancy in the plot saving which I've now attempted to streamline. Simply, if the `-sav` flag is used, the file will be saved, as usual. Otherwise, the figure will appear, again, as usual, and the user can save the Figure directly from the Matplotlib figure interface. This replaces the user input which was getting in trouble with the `plt.show()` commands and saving blank figures.

I also changed the colormap from rainbow to viridis. Hehe.

### DIFF
--- a/programs/forc_diagram.py
+++ b/programs/forc_diagram.py
@@ -14,10 +14,6 @@ import os
 import numpy as np
 import itertools
 import matplotlib
-from matplotlib import pyplot as plt
-import pandas as pd
-import time
-from scipy.interpolate import griddata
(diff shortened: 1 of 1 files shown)


apivarunas:
This PR also fixes #480, I believe.

PR #816, opened 2025-12-05
Add mad_to_a95 conversion function

LenGallo:
Implements conversion from MAD (or aMAD) to a95 using Khokhlov & Hulot (2016, Table 8)


Swanson-Hysell:
The other thing that we need to update is to enable the function to take an array for n_steps values if there is an array of MAD values. It is common with a set of data for there to be different n_steps for different specimens that would have different MADs.

Swanson-Hysell:
I have now added the ability for the function to accept an array of n_steps values that can correspond with the MAD values. It can either accept a single value as before or accept an array of matching length. @LenGallo Can you give the updated function a test and make any additional changes you see fit?

PR #817, opened 2025-12-16
Demag GUI improvements

apivarunas:
Have you ever had the nagging feeling that you dropped a digit from your field notebook to your magnetometer computer? Have you ever wished, while using the Demag GUI, that you could see what orientation information is being used for your data?

Fear not, for this PR adds orientation information to the Demag GUI sidebar. Pretty simple. Does what it says. **Doesn't** provide capability to edit that orientation information directly in the GUI.


Swanson-Hysell:
A small thing is that I always think about Bedding Dip Direction and then Dip (rather than dip then dip direction as presented)
Screenshot 2025-12-16 at 5 03 51 PM

PR #846, opened 2026-04-23
Cleaning Formatting in Demag Gui

apivarunas:
Using recommendations from the `ruff` linter and formatter extension in VS Code, cleaning various issues in Demag Gui. Highlights include:

1. Partially cleaning and re-organizing imports, removing unused imports, cleaning some duplicates.

2. Fixing style of equality comparisons (i.e., `==` to `is`, etc.), type comparisons.

Went from 258 "problems" to 125. Most of what is left is variable assignments, which are a little touchy, bare `except` statements, and some import confusion (mostly from star imports, I think). But, a little at a time and that done well.

### DIFF
--- a/programs/demag_gui.py
+++ b/programs/demag_gui.py
@@ -2873,8 +2873,7 @@ class Demag_GUI(wx.Frame):
         if er_ages_rec["age"] == "":
             if "age_range_high" in list(er_ages_rec.keys()) and "age_range_low" in list(er_ages_rec.keys()):
                 if er_ages_rec["age_range_high"] != "" and er_ages_rec["age_range_low"] != "":
-                    er_ages_rec["age"] = scipy.mean(
-                        [float(er_ages_rec["age_range_high"]), float(er_ages_rec["age_range_low"])])
+                    er_ages_rec["age"] = mean([float(er_ages_rec["age_range_high"]), float(er_ages_rec["age_range_low"])])
         if er_ages_rec["age"] == "":
             return(er_ages_rec)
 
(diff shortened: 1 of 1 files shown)


Swanson-Hysell:
Reviewed the diff and did a smoke test opening a working data directory and
stepping through the GUI — everything behaves as expected.

The changes are carefully largely mechanical (`== None` → `is None`,
`type(x) == Y` → `type(x) is Y`, `== True` → truthy, unused import removal,
`DataFrame`/`Series` → `pd.DataFrame`/`pd.Series`) and I verified each of
the removed imports is no longer referenced in the file. Two changes are
arguably latent bug fixes: `scipy.mean` → `mean` (scipy isn't actually
imported at the top level and `scipy.mean` was removed in SciPy 2.0), and
`math.isnan` → `np.isnan` (math isn't imported either).

The only behavioral tightening is the bare `except:` → `except FileNotFoundError:`
changes in `get_data_info` and `update_pmag_tables`. My smoke test used a
complete working directory so it didn't exercise those paths — worth noting
that `read_magic_file` and `pmag.magic_read` both handle missing files
internally and return `[]` rather than raising, so the `except FileNotFoundError:`
branches will effectively never fire regardless. If anything, it's better
to let unexpected exceptions propagate than silently swallow them, so the
tightening is a net positive.

A couple of small follow-up opportunities spotted while reading:

- Lines 3636 and 3652 have `type(tmin) is str or type(tmin) is str and tmin != ''` —
  the two conditions are identical, so the whole expression reduces to
  `type(tmin) is str`.
- The file has both `import numpy as np` and `from numpy import vstack, sqrt,
  arange, array, mean, nan`, which invites inconsistent `np.array` vs
  `array` usage. Worth consolidating to one style in a separate PR.
- A leftover `#DEBUG` print in `add_fit` (line 2586) that probably makes sense to look into and clean-up.

PR #851, opened 2026-05-05
add recursive-include pmagpy/field_models *.txt to MANIFEST.in

### LINKED ISSUE #850: IGRF14 FileNotFoundError
**What command was run:**

`ipmag.igrf([2000, 1, 19.52, -102.00], mod='arch3k')`

**What happened:**

FileNotFoundError for `igrf14coeffs.txt`

**What I expected to happen:**

I expected the command to output as it used to.

**Data file(s):**

N/A

**PmagPy version:**

`4.3.12`, installed via `pip`. Apparently, it is now `4.3.12` on PyPi, however, the release in the repository is still at `4.3.8`.

I think this must be related to #819. From what I can tell, the folder `field_models` which contains the file `igrf14coeffs.txt` apparently doesn't get made on install. As in, my fresh pip installation of `4.3.12` doesn't have that folder and file. I was able to fix this by adding the folder and file manually, but that's obviously not an ideal workaround. I can look at this tonight, I just wanted to document this behavior.

Swanson-Hysell:
Fixes #850 (I think). `pip install pmagpy` was missing `pmagpy/field_models/igrf14coeffs.txt`, causing `ipmag.igrf()` to raise `FileNotFoundError` for any IGRF14-based call.

PR #819 moved the IGRF14 coefficients into an external data file but did not add the new `pmagpy/field_models/` directory to `MANIFEST.in`. With `include_package_data=True` in `setup.py`, `setuptools` only bundles non-`.py` files that `MANIFEST.in` picks up, so the file was excluded from both the sdist and the wheel uploaded to PyPI. This adds the missing `recursive-include` line, mirroring the existing entry for `pmagpy/data_model/`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 recursive-include pmagpy/data_model *.*
+recursive-include pmagpy/field_models *.txt
 recursive-include images *.txt *.ico
 recursive-include programs/images *.txt *.ico
 recursive-include dialogs/help_files *.html


apivarunas:
Looks good. Thanks for fixing this, @Swanson-Hysell!

PR #855, opened 2026-05-06
fix DataFrame fillna for pandas 3.x compatibility

Swanson-Hysell:
pandas 3.0 raises TypeError when filling NaN in a numeric column with an empty string, breaking the common `df.fillna("", inplace=True)` pattern used throughout the MagIC writer and lab-instrument converters. Replace with `df = df.astype(object).fillna("")` at every call site so the result is consistent across pandas 2.x and 3.x.

Two chained-assignment sites in convert_2_magic.py were already silently broken under pandas 3's copy-on-write rules; the rewrite fixes those too.

Fixes #840.

### DIFF
--- a/dialogs/magic_grid3.py
+++ b/dialogs/magic_grid3.py
@@ -162,7 +162,7 @@ class BaseMagicGrid(gridlib.Grid, gridlabelrenderer.GridWithLabelRenderersMixin)
         Add items and/or update existing items in grid
         """
         # replace "None" values with ""
-        dataframe = dataframe.fillna("")
+        dataframe = dataframe.astype(object).fillna("")
         # remove any columns that shouldn't be shown
         for col in hide_cols:
             if col in dataframe.columns:
@@ -502,7 +502,7 @@ class MagicGrid(BaseMagicGrid):
         Add items and/or update existing items in grid
         """
         # replace "None" values with ""
-        dataframe = dataframe.fillna("")
+        dataframe = dataframe.astype(object).fillna("")
         # remove any columns that shouldn't be shown
         for col in hide_cols:
             if col in dataframe.columns:
@@ -575,7 +575,7 @@ class HugeMagicGrid(BaseMagicGrid):
 
     def add_items(self, dataframe, hide_cols=()):
         # replace "None" values with ""
-        dataframe = dataframe.fillna("")
+        dataframe = dataframe.astype(object).fillna("")
         # remove any columns that shouldn't be shown
         for col in hide_cols:
             if col in dataframe.columns:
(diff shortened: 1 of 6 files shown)


Swanson-Hysell:
## Risks

- `astype(object)` casts every column to object dtype. Most call sites feed straight into `to_dict('records')` or a tab-file write, but any downstream action that does something numerical on the returned DataFrame will see a dtype change.
- Two chained-assignment sites in `convert_2_magic.py` (lines 3602, 3718) were silently a no-op on pandas 3.x; the fix makes them actually replace NaN values, matching pandas 2.x behavior.

## Review testing

- Run the full test suite on both pandas 2.x and 3.x
- Spot-check some lab-instrument converters end-to-end against the example data in `data_files/` and diff the output against what the same call produces on pandas 2.x.
- Run example notebooks (`PmagPy_calculations.ipynb`, `PmagPy_MagIC.ipynb`) on pandas 3.x.

Swanson-Hysell:
## Test results

Tested by running [`MagIC_contribution.ipynb`](https://github.com/Swanson-Hysell-Group/2025_Oman_Paleogeography/blob/main/code/MagIC_contribution.ipynb) (which exercises `ipmag.upload_magic`, `ipmag.combine_magic`, and several `magic_write` calls — i.e. the main fillna call sites this PR touches) end-to-end in two fresh venvs:

| | pandas 2.3.3 | pandas 3.0.2 |
|---|---|---|
| `master` | pass | **fail** — `LossySetitemError` → `TypeError: Invalid value '' for dtype 'float64'` at `pmagpy/ipmag.py:5785` |
| `fix_fillna_pandas_issue` | pass | pass |

Confirms the PR resolves the pandas-3.x regression while remaining compatible with pandas 2.x.

Swanson-Hysell:
## Spot-check: convert_2_magic with example data

Ran four converters touched by the PR end-to-end with each pandas version, using `data_files/`:

- `convert._2g_bin` ← `data_files/convert_2_magic/2g_bin_magic/mn1/mn001-1a.dat`
- `convert._2g_asc` ← `data_files/convert_2_magic/2g_asc_magic/_2g_asc/DR3B.asc`
- `convert.iodp_srm_lore` ← `data_files/iodp_magic/U999A/SRM_archive_data/srmsection_17_5_2019.csv`
- `convert.iodp_dscr_lore` ← `data_files/iodp_magic/U999A/SRM_discrete_data/srmdiscrete_17_5_2019.csv`

All four ran cleanly and produced identical record counts under both pandas 2.3.3 and pandas 3.0.2.

`diff -r` across the two output trees shows a single line difference: in `_2g_bin/measurements.txt`, an integer-valued temperature column stringifies as `273.0` under pandas 2.x and `273` under pandas 3.x. Functionally equivalent (both round-trip to 273 K) but driven by a behavior change in `astype(object)` for integer-valued floats. All other output files (specimens, samples, sites, locations, measurements across the four converters) are identical between the two pandas versions.

Issue #871, opened 2026-05-26
getting rid of distutils in pmagplotlib.py


duserzym:
**What command was run:**

from distutils.version import LooseVersion

**What happened:**

ModuleNotFoundError: No module named 'distutils'

Since Python 3.12, distutils has been removed from the standard library, making its "installation" more about finding its modern replacement: setuptools. 

Can we please remove that?

apivarunas:
Yes, this error broke the automated testing in #875.

EDIT: This should now be fixed with the next commit.

apivarunas:
This was fixed in #842

Issue #769, opened 2025-07-31
pmag_gui does not respond


ElisaUBU:
I have installed pmapy 4.3.7. and pmagcli 4.2.124 via pip on a Windows 11 Enterprise computer, it did not give any errors. When running pmag_gui, it does not respond.

I have also tried to do a vgp calculation. It produces a blank .txt file. And the following errors:

  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\Scripts\dia_vgp.exe\__main__.py", line 4, in <module>
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\site-packages\programs\dia_vgp.py", line 7, in <module>
    import pmagpy.pmag as pmag
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\site-packages\pmagpy\__init__.py", line 23, in <module>
    from . import pmag
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\site-packages\pmagpy\pmag.py", line 17, in <module>
    import SPD.lib.leastsq_jacobian as lib_k
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\site-packages\SPD\__init__.py", line 11, in <module>
    import SPD.spd
  File "C:\Users\emsanchez\AppData\Local\anaconda3\envs\pmagpy_env\lib\site-packages\SPD\spd.py", line 22, in <module>
    from past.utils import old_div
ModuleNotFoundError: No module named 'past'

I'm not a developer, nor do I program in Python, so there are things about the pmagpy community that escape me. But I have been using pmag_gui for many years and have had no problems installing it on different computers and with different versions of Windows. Any idea what might be going on? Thanks!

apivarunas:
Thanks for bringing this up! At first glance it appears as if there is a pesky old_div/ past reference hanging around. Thought we had cleaned all those out, but there were a lot. I'll take a look when I'm back at my computer this afternoon.

apivarunas:
[...]

PR #875, opened 2026-06-20
Thellier GUI fix

apivarunas:
Fixing issue with Thellier GUI that was noticed pursuant to #769. Seems related to #722, but I'm unclear as to why it didn't come up before.


apivarunas:
This also should fix #871 ... I know it's not ideal to do two things at once, but the checks weren't passing.

apivarunas:
Turns out the `distutils` replacement was now already done, so accepted the previous fix.

PR #877, opened 2026-07-06
progress on new coercivity spectra code and hysteresis processing updates

### LINKED ISSUE #876: Wire up unused hysteresis helpers find_hysteresis_turning_point / collapse_hysteresis_field_plateaus
PR #842 added two helpers to `pmagpy/rockmag.py` that are defined but never called:

- `find_hysteresis_turning_point(field)` — finds the single loop reversal while tolerating repeated field plateaus and minor field glitches, choosing the turning point nearest the field extremum.
- `collapse_hysteresis_field_plateaus(field, magnetization)` — averages consecutive repeated field steps into a single point.

Both appear intended to make `split_hysteresis_loop` more robust to imperfect data (the goal of #842), but `split_hysteresis_loop` still uses its own inline `np.gradient`/`np.sign` turning-point detection and falls back to a `n = len(field) // 2` split when more than one turning point is found.

Task: integrate these helpers into `split_hysteresis_loop` (and `grid_hysteresis_loop` as appropriate), or remove them if the crude fallback is preferred. Because this changes branch-splitting behavior on real loops, it should land with test cases covering plateaus, glitches, and multi-turning-point data rather than as a drive-by change.

Found during review of #842.

Swanson-Hysell:
# Coercivity spectrum unmixing and hysteresis processing fixes

This PR contains progress in `pmagpy/rockmag.py` on both coercivity spectrum unmixing and hysteresis loop processing, each with its own test suite. Changes and new features will be documented in rockmagpy notebooks in that separate repository.

## Coercivity spectrum unmixing

A rebuilt coercivity unmixing suite: skew-normal components can be fit either to the coercivity spectrum (MaxUnMix lineage) or directly to the measured backfield/acquisition curve. Includes a method-dispatch entry point (`unmix_coercivity`), bootstrap and Bayesian uncertainty estimation, batch processing of MagIC measurement tables, and export of results to MagIC specimens tables. Tests with known synthetic components are in `pmagpy/test/test_rockmag_unmixing.py`.

## Hysteresis loop processing

The hysteresis processing functions were audited against Jackson & Solheid (2010, doi:10.1029/2009GC002932) — the protocol they implement — and against HystLab (Paterson et al., 2018; MATLAB source consulted directly). 

The changes that were made are associated with:

- **Statistical fixes.** The high-field saturation test mixed sums of squares over different point sets (FNL could go negative; unsaturated loops passed  as saturated, biasing Ms low); `Fnl_lin` now implements eq. 21 of Jackson & Solheid (2010); the loop closure test was inert (classified every loop as closed) and is rebuilt to match HystLab's implementation, including its err(H)-based noise estimate; Q is now computed on the offset-corrected loop (previously a loop offset could depress Q below the Q < 2 gate that decides whether to correct that offset); off-by-one in the prorated drift correction.
- **Q convention retained and documented**: Q = log10(1/sqrt(1−R²)), matching the IRM software and HystLab (the equation as printed in Jackson & Solheid 2010 omits the sqrt; see Paterson et al. 2018, eq. 4).
- **Robust branch splitting** (closes #876): `find_hysteresis_turning_point`(after fixing a latent compressed-index bug) and `collapse_hysteresis_field_plateaus` are now wired into `split_hysteresis_loop`/`grid_hysteresis_loop`, replacing the gradient detection and its half-split fallback. Gridded-loop behavior is unchanged.
- **Stream-agnostic inputs**: new `sanitize_hysteresis_inputs` (lists, numeric strings, NaN dropping with report, non-tesla field warning), and either field sweep order is now handled, so the workflow serves plain field/moment data as well as MagIC tables.
- Docstring updates throughout (including an inverted description of the saturation test result and misdocumented units on `linear_HF_fit`).

Tests: `pmagpy/test/test_rockmag_hysteresis.py` (38 tests on synthetic loops with analytical expectations).

### DIFF
--- a/pmagpy/rockmag.py
+++ b/pmagpy/rockmag.py
@@ -3,6 +3,7 @@ import numpy as np
 import copy
 import json
 import ast
+import warnings
 
 from scipy.optimize import minimize, brent, least_squares, minimize_scalar, brentq
 from scipy.signal import savgol_filter, find_peaks
@@ -6223,13 +6224,23 @@ def _unmix_fit_engine(x, y, initial, method, curve_type='backfield',
     n_pts, n_par = len(x), len(res.x)
     dof = n_pts - n_par
     se = np.full(n_par, np.nan)
+    covariance_ok = True
     if dof > 0:
         try:
             JtJ = res.jac.T @ res.jac
             cov = 2.0 * res.cost / dof * np.linalg.pinv(JtJ)
             se = np.sqrt(np.clip(np.diag(cov), 0, None))
         except np.linalg.LinAlgError:
-            pass
+            # covariance failed on an otherwise-converged fit: warn and flag it
+            # so the resulting NaN standard errors are not mistaken for those of
+            # a degenerate (dof <= 0) fit
+            covariance_ok = False
+            warnings.warn(
+                'covariance matrix could not be computed (singular Jacobian); '
+                'the linearized standard errors are NaN. The parameter '
+                'estimates are still valid -- use bootstrap or Bayesian '
+                'uncertainty for this fit instead.',
+                RuntimeWarning, stacklevel=2)
 
     c, loc, dp, skew, offset = unpack(res.x)
     se_c = se[:K] * y_scale
@@ -6251,6 +6262,7 @@ def _unmix_fit_engine(x, y, initial, method, curve_type='backfield',
         'skew': np.asarray(skew)[order], 'se_skew': np.asarray(se_skew)[order],
         'offset': offset * y_scale, 'se_offset': se_offset,
         'x': x, 'y': y, 'weights': weights, 'y_scale': y_scale,
+        'covariance_ok': covariance_ok,
     }
 
 
@@ -6309,6 +6321,10 @@ def _build_unmix_result(engine, method, curve_type, vary_skew, fit_offset,
         'reduced_chi_square': rss / dof if dof > 0 else np.nan,
         'aic': n * log_term + 2 * n_params,
         'bic': n * log_term + n_params * np.log(n),
+        # True when the covariance/standard-error computation failed (singular
+        # Jacobian): the se_* fields are NaN for that reason rather than
+        # because the fit was degenerate. Normally False.
+        'covariance_singular': not engine.get('covariance_ok', True),
     }
 
     scipy_result = engine['scipy_result']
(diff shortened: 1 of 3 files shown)


duserzym:
Review of the Review and fixes to the hysteresis processing code in rockmag.py markdown document:

We can merge changes and test the hysteresis notebooks. 

Item 1: I confirm that the Q factor calculation remains untouched. I remember finding this issue last year in the original 2010 paper. Although the hystlab paper pointed out the problem already, it is probably worth mentioning the correction again in the new manuscript. 

Item 2: change is correct and should be accepted.

Item 3: good catch.

Item 4: I was wrong, my bad! The IRMDB function had it right. 

Item 5: good catch. 

Item 6: correct. 

The lingering issue item 11 removing the Qf mode computation for calc_Q calculation is good, since that function should aim at calculating general quality with any give field-moment pair.

Swanson-Hysell:
Here are some more details about this PR which adds a coercivity spectrum unmixing framework to `rockmag.py` and corrects the hysteresis loop processing workflow.

**New unmixing API.** Skew-normal coercivity components fit in spectrum or measurement space through a shared least-squares engine, with a method registry (`unmix_coercivity`; `spectrum`, `curve`, `maxunmix`, `bayes`), parsimony-based model-count selection (`select_n_components`), multistart solution mapping (`unmixing_multistart`), bootstrap uncertainties (`unmixing_bootstrap`), a faithful MAX UnMix reproduction, and Bayesian unmixing via nested sampling with a mineral-prior library (`mineral_priors`, mean-coercivity parameterization). Batch processing (`unmix_backfield_experiments`), count-robust class aggregation (`aggregate_by_class`), plotting, and MagIC specimens-table export (`rem_cmf`/`rem_cd` rows or JSON description payloads) are included.

**Hysteresis processing corrections** (Jackson & Solheid, 2010): corrected Q, SSPE windowing, and F-statistics; drift correction now applied in true measurement-time order for loops measured from either saturation; robust behavior on open, unsaturated, and sparse loops (NaN + warning instead of crashes). Values of Q/FNL from earlier versions are not comparable.

**Behavior changes to note:**
- Unified defaults across the unmixing entry points: `method='spectrum'`, `vary_skew=True` (`unmix_coercivity_bayes` deliberately keeps `vary_skew=False`).
- Specimens `description` cells now use a `'free text | JSON'` convention shared by the hysteresis and unmixing writers; legacy Python-dict cells are still read and migrated.
- `backfield_MaxUnmix` is deprecated (FutureWarning) in favor of `unmix_coercivity(method='maxunmix')`.
- `loop_closure_test` keeps its original `(H, Mrh, HF_cutoff)` positional signature; new parameters are keyword-only.

PR #856, opened 2026-05-07
Removal of rock mag functions from ipmag

apivarunas:
This PR deals with #802. The main difficulty here was dealing with the CLI programs which were wrappers on the `ipmag` functions. In particular, it was difficult to implement `hysteresis_magic.py` (the command line `ipmag` wrapper). Of particular note, it seems as if the calculations from `rockmag` functions are somewhat different than what is getting implemented from `pmagplotlib`, which I've gotten back to. There was also some oddities in how `rockmag` functions ended up slope-correcting some of the example functions.

Something to look into.

Issue #802, opened 2025-10-16
Removing rock magnetic functions now in rockmag


apivarunas:
There are some functions in `ipmag.py` that are now implemented in `rockmag.py`:

`dayplot_magic`
`smooth`
`curie`
`iplot_hys`
`hysteresis_magic`

To avoid redundancy, I suggest that these functions be removed and their implementations stay in the `rockmag` module, which is good practice anyway in order to create more focused modules.

apivarunas:
Alright. This is what has been done on this.

`curie` has been removed from `ipmag`. `curie.py`, the CLI program, has been rewritten to use the `rockmag` functions.

`smooth` has been removed. It was only used in `curie`.

`hysteresis_magic` has been removed from `ipmag`. The `hysteresis_magic.py` CLI program has been rewritten to use the `rockmag` functions. And `hysteresis_magic2.py` has been removed, as per #789.

`dayplot_magic` has been removed from `ipmag`. The `dayplot_magic` CLI program has been rewritten to use the `rockmag` functions. `dayplot_magic2` has been removed, as per #789.

`iplot_hys` has been removed from `ipmag`. It was only used in `ipmag`.

`plot_day`, `plot_irm` have been removed from `pmagplotlib.py`. Will wait for another time to fix `plot_hys` due to some dependencies.

apivarunas:
Okay, I'm having more trouble than I might wish testing out these changes. Basically all in the CLI implementation. `rockmag` simply does not play nice with current organization.

The inconsistent column naming (i.e., `hyst_mr_moment` as column name for current CLI example files versus defaults of `hyst_mr_mass`, in `rockmag`, as just one example) is a consistent issue that is going to be a pain to resolve. 

@Swanson-Hysell Do we want to keep all the CLI programs?

Swanson-Hysell:
On the Curie-temperature piece of this, I think we're ready for the consolidation (after we merge in #878).

[...]

PR #878, opened 2026-07-09
Add multi-method Curie temperature estimation to rockmag.py

Swanson-Hysell:
## Summary

This PR adds multi-method Curie temperature estimation to `pmagpy/rockmag.py`, designed so that the method-dependent biases among estimators are explicit rather than hidden behind a single reported number. It builds on the initial Curie functionality from #776, which it refactors and supersedes.

Curie temperature estimates from thermomagnetic data are method- and data-type-dependent, with systematic offsets of several to tens of degrees documented on synthetic titanomagnetites ([Lattard et al., 2006](https://doi.org/10.1029/2006JB004591)). The Landau-theory analysis of [Fabian et al. (2013)](https://doi.org/10.1029/2012GC004440) shows that for in-field M(T) curves the Curie temperature corresponds to the inflection point (minimum of dM/dT), while the classical maximum-curvature (Ade-Hall et al., 1965) and two-tangent ([Grommé et al., 1969](https://doi.org/10.1029/JB074i022p05277)) estimates lie systematically above it; for low-field χ(T) curves the two-tangent construction is not physically justified and the Curie-Weiss extrapolation of 1/χ yields the paramagnetic Curie temperature θ ≥ Tc ([Petrovský & Kapička, 2006](https://doi.org/10.1029/2006JB004507)). The implementation attaches these caveats to its outputs.

## What's added

- `curie_temperature_estimates()` — main entry point: applies selected methods to the heating and cooling branches of a MagIC-formatted experiment and returns a tidy comparison table (one row per branch × method) with per-method caveat notes. A `data_type` parameter (`'susceptibility'`/`'magnetization'`, inferred from the column name by default) controls the default method set and caveats; `method_kwargs` keys are validated so typos raise rather than being silently ignored.
- Per-method estimators operating on plain temperature/magnetization arrays (so non-MagIC instrument exports can be analyzed directly):
  - `curie_derivative_estimates()` — inflection point and maximum-curvature estimates, with the curvature search restricted to the transition zone
  - `curie_two_tangent()` — intersecting tangents with auto or explicit windows; flags the case where no flat baseline exists (curve ending below Tc)
  - `curie_inverse_susceptibility()` — reproducible Curie-Weiss fit over a stated window with r², Curie constant, and propagated 1σ on θ; rejects non-positive slopes and warns when the window contains resolution-limited (repeated identical) susceptibility values, which bias θ low
  - `curie_landau_fit()` — fit of the in-field Landau equation of state m³ + τm = h (scipy-only; no new dependencies) with formal uncertainty; automatically flags Moskowitz (1981)-style extrapolation when data end below Tc, where the inflated uncertainty makes the non-identifiability explicit
- `landau_magnetization()` — closed-form (Cardano/trigonometric) solution of the Landau equation of state
- `plot_curie_estimates()` — multi-panel comparison figure (curve + per-method Tc lines + tangent/Landau constructions, derivative panel, 1/χ panel); consumes the same estimator results as the table (single computation, single source of truth)
- `interactive_curie_inverse_susceptibility()` — the draggable Bokeh Curie-Weiss fit, now a standalone `interactive_*` function
- `add_curie_estimates_to_specimens_table()` — writes `critical_temp` (in K) and `critical_temp_type` to a MagIC specimens table with the estimation method, branch, and uncertainty archived in `description`. Existing description content is preserved (dict literals updated, free text kept and appended to), and experiment matching is exact or colon-token based — never substring — so estimates cannot be written to the wrong specimen's rows.
- `prepare_thermomag_branches()` — shared preprocessing: branch splitting, optional holder correction, temperature-space smoothing, and duplicate-temperature averaging (repeated furnace-stabilization readings otherwise over-weight fits and produce zero-spacing derivatives).

## What's changed

- `split_warm_cool()` rewritten on numpy: filters non-finite rows, returns arrays, and handles cooling-only runs (the first point takes the direction of the first step instead of always being assigned to heating).
- `smooth_moving_avg()` no longer returns NaN when a tapered window degenerates to ≤2 points.
- `plot_X_T()` accepts Kelvin output, tolerates single-branch experiments, and averages duplicate temperatures before differencing (fixes inf/NaN derivative traces on real Kappabridge data).

## What's removed

- `estimate_curie_temperature()` (added in #776; no callers in this repo or in RockmagPy-notebooks). Its derivative-based estimates map onto the `inflection` and `max_curvature` methods of the new API. The legacy `ipmag.curie` second-derivative value corresponds to `max_curvature`: on `data_files/curie/curie_example.dat` with a 10-degree window, `ipmag.curie` gives 552°C and `max_curvature` gives 549°C (the small difference reflects the different smoothing kernels).

## Relationship to #856 / #802

This PR is complementary to the consolidation of rock magnetic analysis in `rockmag.py` (#856). Note that #856 currently rewrites `programs/curie.py` as a plot-only wrapper, dropping the CLI's quantitative Tc output; once this PR lands, `programs/curie.py` could wrap `curie_temperature_estimates()` (with `max_curvature` as the legacy-compatible default and a `--method` flag) to restore it.

## Testing

New `pmagpy/test/test_rockmag_curie.py` with 56 tests against synthetic curves with known analytical answers: exact recovery of Tc by the Landau fit and of θ by the Curie-Weiss fit; the documented bias ordering (two-tangent ≈ max curvature > inflection ≈ Tc) on in-field Landau curves; extrapolation-mode flagging and uncertainty inflation; resolution-limit detection; duplicate-temperature, heating-only, and holder-correction handling; MagIC specimens-table round-trips (Kelvin conversion, controlled vocabulary, description preservation, matching); and a regression lock against the legacy `ipmag.curie` value on `curie_example.dat`. Full suite: 452 passed.

Companion documentation (Jupyter Book notebook demonstrating all methods and caveats on Rock Magnetic Bestiary data, plus a written evaluation of the legacy Curie code) is in RockmagPy-notebooks.

## Notes for reviewers

- `remove_holder=True` (per-branch minimum subtraction) remains the default; the docstring documents that this assumes the run passes above Tc, and the notebooks demonstrate `remove_holder=False` for runs that end below it (e.g., MPMS M(T) curves).
- Derivative-based values differ slightly from legacy `ipmag.curie` by design (transition-zone restriction, interpolated zero crossings); see the mapping above.

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -9,6 +9,7 @@ import re
 import sys
 import time
 import urllib
+import warnings
 import zipfile
 import io
 
@@ -8338,6 +8339,22 @@ def curie(path_to_file='.', file_name='', magic=False,
     The estimated curie temperation is the maximum of the 2nd derivative.
     Temperature steps should be in multiples of 1.0 degrees.
 
+    .. deprecated::
+        ``ipmag.curie`` is deprecated and will be removed in a future release.
+        It reports a single Curie temperature from the maximum of the smoothed
+        second derivative. Use the multi-method estimators in ``pmagpy.rockmag``
+        instead, which make method-dependent biases explicit (Fabian et al.,
+        2013, doi:10.1029/2012GC004440):
+
+        - ``rockmag.curie_temperature_estimates()`` applies the selected methods
+          to the heating/cooling branches of a MagIC experiment and returns a
+          tidy comparison table with per-method caveats.
+        - ``rockmag.curie_derivative_estimates()`` is the direct analog of this
+          function, returning both the inflection-point and maximum-curvature
+          estimates (the ``max_curvature`` method reproduces the legacy value
+          here: 552 C vs 549 C on data_files/curie/curie_example.dat with a
+          10-degree window).
+
     Parameters:
         file_name : name of file to be opened
         path_to_file : path to directory that contains file (default is current directory, '.')
@@ -8352,6 +8369,14 @@ def curie(path_to_file='.', file_name='', magic=False,
     Returns:
         A plot is shown and saved if save=True.
     """
+    warnings.warn(
+        "ipmag.curie is deprecated and will be removed in a future release. "
+        "Use pmagpy.rockmag.curie_temperature_estimates (MagIC experiments) or "
+        "pmagpy.rockmag.curie_derivative_estimates (the max_curvature/inflection "
+        "analog of this function) instead.",
+        FutureWarning,
+        stacklevel=2,
+    )
     plot = 0
     window_len = window_length
 
--- a/pmagpy/rockmag.py
+++ b/pmagpy/rockmag.py
@@ -4461,6 +4461,23 @@ def _inverse_susceptibility(chi):
     return np.divide(1.0, chi, out=np.full_like(chi, np.nan), where=chi > 0)
 
 
+def _repeated_value_fraction(values):
+    """
+    Fraction of points that repeat an identical neighbor.
+
+    Runs of identical values within a fitting window are the signature of a
+    resolution-limited (quantized) susceptibility tail or a flattened/altered
+    magnetization tail; either flattens a fitted slope and biases the
+    extrapolated Curie temperature. Returns a value in [0, 1].
+    """
+    values = np.asarray(values, dtype=float)
+    if values.size < 2:
+        return 0.0
+    repeated = np.concatenate([[False], values[1:] == values[:-1]])
+    repeated[:-1] |= repeated[1:]
+    return float(repeated.mean())
+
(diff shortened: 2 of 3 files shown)


apivarunas:
Wow! That's a lot of enhancement.

Swanson-Hysell:
Here is are some plots of the new code in action:
Screenshot 2026-07-09 at 10 28 54 AM

Swanson-Hysell:
Just bumped into an issue associated with the inflection method picking up a subsidiary trough in dM/dT:
Screenshot 2026-07-09 at 12 38 12 PM

Swanson-Hysell:
These functions can still be improved moving forward, but I am going to merge in for now so that the rockmagpy notebooks that use these functions can run.

Issue #825, opened 2026-03-01
Add `random_seed` parameter to svei.py random sampling functions


Swanson-Hysell:
### Context

As part of the reproducible random sampling effort (#447), all scientific functions in `pmag.py` and `ipmag.py` that use random number generation are being converted to accept an optional `random_seed` parameter using the `numpy.random.default_rng()` API via `pmag._resolve_rng()`.

The `svei.py` module (secular variation of the geomagnetic field — GGP models, Anderson-Darling tests) was not included in that initial conversion but contains functions that use `np.random.multivariate_normal` for Monte Carlo sampling.

### Functions requiring conversion

**Direct random calls:**

1. **`GGPrand`** (line ~1376) — generates random directions from a GGP model
   - `np.random.multivariate_normal(m, Cov, n)` 
   - Called by `svei_test` in its Monte Carlo simulation loop

2. **`GGP_vMF_cdfs`** (line ~1383) — generates CDFs from GGP + von Mises-Fisher distributions
   - `np.random.multivariate_normal(m, Cov, n)` (GGP vectors, n=2E6 default)
   - `np.random.multivariate_normal([0,0,0], C, n)` (vMF deviations)
   - Called by `AD_test_w_kappa` and `svei_test`

**Callers needing `random_seed` propagation:**

3. **`svei_test`** (line ~2216) — calls `GGPrand`, `GGP_vMF_cdfs`, and `ipmag.fishrot`
4. **`AD_test_w_kappa`** (line ~1708) — calls `GGP_vMF_cdfs`

### Implementation pattern

Follow the same pattern used in `pmag.py` and `ipmag.py`:

```python
def GGPrand(GGPmodel, lat, n, degree=8, random_seed=None):
    rng = pmag._resolve_rng(random_seed)
    m = m_TAF(GGPmodel, lat)
    Cov = Cov_modelo(GGPmodel, lat, degree)
    X = rng.multivariate_normal(m, Cov, n) * 1000
    DI = pmag.cart2dir(X)
    return DI
```
Callers pass random_seed=rng to propagate the Generator downstream.

####  Notes
GGP_vMF_cdfs draws 2 million samples by default — reproducibility is particularly valuable here for testing and validation
svei_test also calls ipmag.fishrot (already converted) at line ~2348, so random_seed propagation to that call should be included
numpy.random.Generator.multivariate_normal is a direct replacement for np.random.multivariate_normal with the same signature

PR #879, opened 2026-07-15
Accelerate find_flat simulations and fix figure export

duserzym:
## Summary

- batch Gaussian field and Fisher resampling across simulations
- compute orientation matrices and eigensystems with vectorized NumPy operations
- replace repeated full-sample CDF scans with equivalent integer-degree bin counts
- reuse the initial factor-1 result instead of running it twice
- make `save=True` reliably write the SVEI diagnostic and final summary figures
- fix the exported `V2max` column, safe output naming, and a plotting string error
- add regression coverage for numerical equivalence, batched outputs, and figure export

## Root cause

`find_flat` called `GGPrand` once per simulation and, with finite kappa, called `fishrot` and `fisher_mean` once per direction inside every simulation. Each flattening factor also constructed marginal CDFs by scanning two million sampled directions once for every degree-grid point.

## Impact

On a single-thread local benchmark with 222 directions, 70 flattening factors, 1,000 simulations per factor, and two million CDF samples per factor, the complete scan finished in 42.5 seconds. Before this change, the same workload was estimated at roughly 29 minutes from component timings.

The defaults and sampling model are unchanged. CDF grid values match the former implementation to floating-point precision, and the no-kappa batched Monte Carlo results match the former scalar loop under the same random seed.

## Validation

- `MPLBACKEND=Agg pytest -q` — 400 passed
- production-default 70-factor end-to-end benchmark
- saved quick scan verified both the summary and `_svei_test` PDF outputs


Swanson-Hysell:
Hi @ltauxe Might you be able to take a look here since this pull request makes changes to  svei.py?

Two things going on. The big one is speeding up find_flat — the Monte Carlo draws and the CDF building are now vectorized and batched instead of looping one simulation at a time. @duserzym describes the big resulting speed up above. The other is adding a random_seed argument through the functions so you that there is reproducibility of a single run (issue #825 and something we have implemented elsewhere). If you don't pass a seed, it behaves the same.

A couple other little bugs are fixer here. The V2max column in the find_flat output
was getting filled with the V2min values and there was an issue with figure saving.

The numbers don't change as far as I can tell. The distributions of best flattening factor, A2I, A2D, E and H all come out the same. Elongation is identical, which makes sense since it's set by the data. SVEI_demo.ipynb runs start to finish without errors.

PR #886, opened 2026-07-19
Fix matplotlib warnings in ipmag.fishqq

### LINKED ISSUE #824: Matplotlib warnings in ipmag.fishqq()
`ipmag.fishqq()` triggers two matplotlib warnings when called in a session where figures already exist (e.g., during testing):

**1. Figure reuse warning** (`ipmag.py` line 2060):

```
UserWarning: Ignoring specified arguments in this call because figure with num: 1 already exists
```

`plt.figure(fignum, figsize=(6, 3))` passes `figsize` when requesting an existing figure number. Matplotlib ignores `figsize` for already-created figures and warns.

**Fix:** Either call `plt.close(fignum)` before creating, or use `fig = plt.figure(fignum); fig.set_size_inches(6, 3)` to resize an existing figure.

**2. tight_layout warning** (`ipmag.py` line 2066):

```
UserWarning: tight_layout not applied: number of columns in subplot specifications must be multiples of one another
```

`plt.tight_layout()` is incompatible with the subplot layout created by `plot_qq_unf` and `plot_qq_exp`. This means the layout adjustment is silently skipped.

**Fix:** Either use `fig.subplots_adjust(...)` with explicit spacing, or switch to `constrained_layout=True` in the figure constructor.

Both warnings are cosmetic — `fishqq()` produces correct results — but they indicate the figure sizing and layout are not being applied as intended.

Swanson-Hysell:
Clear any existing figure before drawing QQ subplots so the figure size is applied and stale axes don't break tight_layout. Closes #824

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -2113,7 +2113,11 @@ def fishqq(lon=None, lat=None, di_block=None,plot=True,save=False,fmt='png',save
         Dtit = 'Mode 1 ' + dec_label
         Itit = 'Mode 1 ' + inc_label
         if plot:
-            plt.figure(fignum,figsize=(6, 3))
+            # clear any existing figure with this number so that the figure
+            # size is applied and stale axes do not break tight_layout
+            fig = plt.figure(fignum)
+            fig.clear()
+            fig.set_size_inches(6, 3)
             fignum+=1
         else:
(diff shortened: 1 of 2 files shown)


Swanson-Hysell:
Looks good in a test run. Going to merge.
Screenshot 2026-07-19 at 9 24 24 AM

PR #888, opened 2026-07-19
forc processing

duserzym:
added a forc.py file based on FORCme functions but follows the FORCinel and VARIFORC programs that are published. This is different from the forc_diagram.py.

Demo notebook is at https://github.com/PmagPy/RockmagPy-notebooks/pull/67.


Swanson-Hysell:
Given that the FORCme code was merged in as forc.py, it makes sense to close this functionality without merging. More testing and enhancement on that code would be welcome!

PR #890, opened 2026-07-22
Speed up Pmag GUI startup and add platform-specific macOS/Windows builds

duserzym:
## Summary

This updates the Pmag GUI standalone build workflow for both Apple-silicon macOS and 64-bit Windows, while substantially reducing the time required to display the main GUI window.

The branch is synchronized with PmagPy 4.4.0 and replaces the former shared PyInstaller configuration with explicit platform-specific specs and reproducible minimal build environments.

## Why this change is needed

The frozen application previously initialized a large portion of PmagPy before showing its main window. Importing PmagPy, the dialog package, Demag GUI, Thellier GUI, menu dialogs, notebook support, GIS libraries, and multiple plotting backends up front caused startup to take minutes in some packaged builds.

The same shared spec was also being used across operating systems even though macOS bundles and Windows executables have different architecture, runtime-library, signing, and packaging requirements.

## Startup changes

- Load `pmagpy` and `dialogs` submodules on demand through package-level lazy accessors.
- Defer Demag GUI, Thellier GUI, metadata, conversion, orientation, upload/download, and menu-dialog imports until the corresponding action is selected.
- Avoid importing IPython/Jupyter support in frozen applications.
- Keep Cartopy optional and dynamically imported so it does not pull the GIS stack into lightweight frozen builds.
- Retain a startup probe activated only by `PMAG_GUI_STARTUP_PROBE_FILE`, allowing repeatable end-to-end timing through the main frame's `Show()` call.
- Use one-directory PyInstaller layouts so startup does not pay the extraction cost of a one-file bundle.

This shifts specialized initialization to the first time a user opens the relevant tool. The main launcher becomes responsive much sooner, and the existing busy indicator remains visible while a deferred analysis GUI loads.

### macOS / Apple silicon

- Rename the maintained macOS spec to `pmag_gui_macos_arm64.spec`.
- Require a native Apple-silicon build host and explicitly emit `arm64` Mach-O binaries.
- Build a macOS `.app` bundle using the optimized one-directory layout.
- This spec supports Apple M-series Macs only; it does not produce an Intel or universal application.

### Windows

- Add `pmag_gui_windows.spec` for 64-bit Windows builds.
- Add `setup_scripts/pmag_gui_windows_compile.yml` with the tested minimal dependency set.
- Produce a one-directory Windows application and document that the adjacent `_internal` directory must be distributed with the executable.
- Select optional SciPy imports only when present.
- Prefer Intel runtime DLLs from the active Conda environment, preventing SciPy extensions from binding to ABI-incompatible DLLs left on `PATH` by another Conda installation.
- Disable UPX for the Windows build to avoid additional startup and binary-compatibility variability.

The existing macOS environment and standalone-build documentation are updated to use the explicit platform specs. Linux documentation now states that a dedicated Linux spec is still required.

## Measured impact

On the Apple-silicon build machine, five frozen launches of Pmag GUI 4.4.0 reached the main window in:

- 3.838 seconds on the first measured launch
- 2.863, 2.855, 2.845, and 2.885 seconds on subsequent launches
- **2.863-second median**, compared with **11.866 seconds** before the additional lazy-loading work (approximately 76% faster)

## Validation

- Built the macOS application with PyInstaller 6.20.0 from the minimal arm64 environment.
- Confirmed the frozen app launches and reports `pmagpy-4.4.0`.
- Confirmed all 272 Mach-O files in the bundle are arm64-only, with no Intel or universal slices.
- Confirmed the app passes strict deep code-signature integrity verification with PyInstaller's ad-hoc signature.
- Confirmed deferred Demag GUI, Thellier GUI, and dialog modules remain present in the PyInstaller archive and import successfully.
- Confirmed normal non-frozen Cartopy imports still work after the dynamic-import changes.
- Python source compilation and `git diff --check` pass.
- GitHub Actions test jobs pass on Python 3.10 and 3.12.

The Windows spec cannot be executed on the macOS build host, so the resulting Windows application should receive a final build-and-launch smoke test on 64-bit Windows before its standalone release is published.

## User-visible limitations

- The macOS build is Apple-silicon-only.
- The Windows build requires 64-bit Windows and must be distributed as the complete one-directory package.
- Cartopy and its GIS dependency stack are intentionally omitted from these lightweight frozen builds, so optional mapping features remain unavailable there.
- The macOS artifact is ad-hoc signed unless a Developer ID signing/notarization step is performed separately.


apivarunas:
@duserzym I can try running this on a Windows machine if that's helpful!

duserzym:
Hey @apivarunas that would be great! Thanks! I just did some light testing with a lab computer here. I am doing this release because we are running the summer school now and some machines have issues. Hopefully this works ok...!

Thanks again!

PR #897, opened 2026-07-29
update inclination only calculation to include the Arason and Levi (2010) method

Swanson-Hysell:
This pull request introduces improvements to the inclination-only mean calculation routine `doincfish` in `pmagpy/pmag.py`, adding a new estimation method.

**Enhancements to inclination-only statistics:**

* Added support for the Arason and Levi (2010) maximum likelihood estimator to `doincfish`, selectable via a new `method` parameter. This method is robust for steep data and provides profile-likelihood confidence intervals. [[1]](diffhunk://#diff-4449e70e4f385c85aa1e7a4859d4c54ae6bc538c24cdee16379e4ff5fcb6f058L6240-R6260) [[2]](diffhunk://#diff-4449e70e4f385c85aa1e7a4859d4c54ae6bc538c24cdee16379e4ff5fcb6f058R6359-R6450)
* Improved the fallback behavior in the McFadden and Reid method when no root is found, clarifying the warning message and ensuring robust handling of both array and scalar indices.
* Updated the docstring and return dictionary of `doincfish` to document the new method and its additional confidence interval outputs. [[1]](diffhunk://#diff-4449e70e4f385c85aa1e7a4859d4c54ae6bc538c24cdee16379e4ff5fcb6f058L6240-R6260) [[2]](diffhunk://#diff-4449e70e4f385c85aa1e7a4859d4c54ae6bc538c24cdee16379e4ff5fcb6f058R6274-R6278)

**Testing and validation:**

* Added a new test suite (`test_inclination_only.py`) to validate both estimation methods, covering published examples, cross-method agreement, synthetic data recovery, and error handling for invalid method names.

**API improvements:**

* Added a new parameter `iconf` to `plot_aniso` in `ipmag.py`, allowing suppression of the mean-tensor/confidence-ellipse plot for unoriented cores. [[1]](diffhunk://#diff-1880d832ee3c65857bad94da95e2403f5b85ae47a955b5cda2d08dfaeab19464L11607-R11607) [[2]](diffhunk://#diff-1880d832ee3c65857bad94da95e2403f5b85ae47a955b5cda2d08dfaeab19464R11632-R11635)

### DIFF
--- a/pmagpy/ipmag.py
+++ b/pmagpy/ipmag.py
@@ -11604,7 +11604,63 @@ def plot_gc(poles, color='g', fignum=1):
 
 
 def plot_aniso(fignum, aniso_df, Dir=[], PDir=[], ipar=False, ihext=True, ivec=False,
-               iboot=False, vec=0, num_bootstraps=1000, title="", iconf=True):
+               iboot=False, vec=0, num_bootstraps=1000, title="", plot_mean=True):
+    """
+    Plot anisotropy eigenvectors and optional mean-tensor confidence estimates.
+
+    The first figure (fignum) shows the individual specimen eigenvectors
+    (V1 squares, V2 triangles, V3 circles) on an equal-area net. When
+    plot_mean is True, a second figure (fignum+1) shows the eigenvectors of
+    the mean tensor with confidence estimates: Hext ellipses and/or
+    bootstrapped eigenvectors or bootstrap ellipses, with additional
+    figures for bootstrap eigenvalue and eigenvector-component CDFs as
+    requested.
+
+    Parameters
+    ----------
+    fignum : int
+        matplotlib figure number for the eigenvector plot; the mean-tensor
+        figure uses fignum+1 and bootstrap CDF figures use fignum+2 onward
+    aniso_df : pandas DataFrame
+        anisotropy data with an 'aniso_s' column of colon-delimited
+        six-element tensor strings (MagIC data model format)
+    Dir : list
+        [declination, inclination] of a comparison direction plotted on the
+        mean-tensor figure and, with iboot and ivec, compared against the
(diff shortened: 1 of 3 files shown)


Swanson-Hysell:
### Removing the shallow-inclination guard from `doincfish`

Looking further at the McFadden & Reid method turned up related issues in the pre-existing code:

1. The `MI < 30` guard returned the Gaussian mean padded with zero-filled statistics (`k=0`, `csd=0`, zero confidence limits) and omitted the `alpha95` key. The zeros are a bit confusing (a `csd` of 0 reads as perfect data).

2. The guard was masking a latent bug in the root selection. For shallow data the McFadden & Reid fitness function has two roots — the true likelihood maximum plus a spurious near-flat root around 63–67° — and the curvature test `U` was broadcast from shapes `(m,1) − (m,)` into an `(m,m)` matrix indexed with a flattened `argmin`.

3. Checking the root selection against the original paper revealed that `U` also used sec²θ̂₀ where McFadden & Reid (1982, eq. 19a) specify cosec²θ̂₀. With the correct formula, the true likelihood maximum is the unique root with negative `U` and the spurious roots are positive, as the paper states; with the sec² version all roots tested negative, so the selection was only accidentally correct.

The estimator itself is valid for shallow data: Arason & Levi (2010) note the modified McFadden & Reid method gives reasonable results for inclinations below 60° (κ > 10). Testing against synthetic Fisher-distributed data with true inclinations of 5–25° (κ = 10–50, n = 8–50) confirms that, once `U` is computed as a length-m vector with the eq. 19a formula, the curvature criterion uniquely identifies the true root in each case (e.g. U = +99 vs −1662 for the spurious and true roots) and the result agrees with the Arason–Levi MLE to within 0.05°. The McFadden & Reid appendix numerical example reproduces unchanged.

PR #903, opened 2026-08-01
Fix resample count and ellipse hemisphere in find_compilation_kent

Swanson-Hysell:
The Fisher resampling loop in `find_compilation_kent` iterated over the length of the compilation of f factors rather than over `n`, the number of requested resamples. With the default 70-value compilation this silently ignored `n`: values above 70 undersampled the f distribution, and values below 70 raised an `IndexError` when indexing `compilation_incs`. Iterating over `range(n)` produces one Fisher resample of `n_fish` mean poles per resampled f factor, for `n * n_fish` resampled mean poles in total, which is the behavior the docstring describes.

The Kent confidence ellipse is now drawn in the hemisphere of the mean pole itself. `pmagplotlib.plot_ell` returns the lower-hemisphere (stereonet) projection, which for a southern-latitude mean pole is the antipode, so the mean symbol and its ellipse were plotted on opposite sides of the globe. `plot_pole_ellipse` now defaults to `lower=False`, which also fixes `find_ei_kent` and `find_svei_kent`. Northern-hemisphere mean poles are unchanged.

### Changes

- The f compilation is now `ipmag.PIERCE2022_F_COMPILATION` at module level, so the default can be inspected and another supplied via `f_from_compilation`.
- `n` is used in the resampling loop. Note on the default: the old signature said `n=10000`, but the loop never used `n`, so every call produced 70 × `n_fish` = 7,000 mean poles. The new default `n=1000` gives 100,000 mean poles (about 6 s per call); keeping `n=10000` would now mean 1,000,000 poles (about 60 s). The change is therefore a tenfold increase in resampling relative to what actually ran before, not a reduction.
- `plot_pole_ellipse` defaults to `lower=False`. **This is a public default change.** Calls with northern-hemisphere means render identically. Calls that passed a southern-hemisphere Kent mean previously drew the ellipse at the antipode and will now draw it around the mean; external notebooks that relied on the old placement, or that passed `lower=True` explicitly to work around it, should be checked. The parameter remains available for the stereonet convention.
- `plot_pole_ellipse` docstring corrected. It listed `legend`/`filled_pole`/`fill_color`/`fill_alpha` carried over from `plot_pole` rather than its own `edgecolor`/`alpha`/`lw`, and described the default square marker as a circle.

### Tests

`pmagpy/test/test_ipmag_compilation_kent.py` (new, 9 tests): `n` honored below and above the compilation length, a user-supplied compilation, the Kent mean staying in the pole's hemisphere, `PIERCE2022_F_COMPILATION` length and range, and the ellipse hemisphere at both the `plot_ell` and `plot_pole_ellipse` levels, including that the two `lower` choices are antipodal point sets. Six of the nine fail against master. Map-dependent tests skip when cartopy is not installed. Full active suite: 784 passed.

Verification details are in the review comment below.


Swanson-Hysell:
Changes since this PR was opened:

- The f compilation is now `ipmag.PIERCE2022_F_COMPILATION` at module level, so the default can be inspected and another supplied via `f_from_compilation`.
- `n` is now used in the resampling loop, which previously ran once per compilation entry (70) regardless of `n`, for 7,000 mean poles. The default is set to 1000, giving `n * n_fish` = 100,000 mean poles (~6 s per call); the old 10000 default was never exercised by the loop and would now mean 1,000,000 (~60 s).
- The hemisphere fix moved from the `find_compilation_kent` call site into `plot_pole_ellipse` itself (`lower=False` default), so `find_ei_kent` and `find_svei_kent` are fixed too.
- `plot_pole_ellipse` docstring corrected — it listed `legend`/`filled_pole`/`fill_color`/`fill_alpha` carried over from `plot_pole` rather than its own `edgecolor`/`alpha`/`lw`, and described the default square marker as a circle.

Swanson-Hysell:
## Review (2026-09-19)

Both fixes verified by running the branch code; the PR is ready to merge.

### Verified

- **Resample count.** `compilation_incs` is built from `n` draws of the f compilation, so iterating over `range(n)` is the only consistent choice. On the branch, `n=50, n_fish=10` returns 500 mean poles where master raised `IndexError`, and `n=80` returns 80 × n_fish poles where master silently used 70. The default (`n=1000, n_fish=100`) produces 100,000 poles in ~6 s, as described above.
- **Ellipse hemisphere.** For a Kent mean at 278.8°E, 53.7°S, `pmagplotlib.plot_ell(..., lower=True)` returns the ellipse at +44° to +62° near 100°E (the antipode) and `lower=False` returns −62° to −44° at 280°E, on the mean. `plot_ell` with `lower=True` reflects the center and both axes to positive inclination, which is the stereonet convention and not what a map wants. `pmag.dokent` does no flipping of its own, so nothing upstream compensated; the mean symbol was already plotted at the pole, so the old default put symbol and ellipse in opposite hemispheres.
- **Callers.** The three internal callers (`find_ei_kent`, `find_compilation_kent`, `find_svei_kent`) all use the default and are fixed by it. No other callers in the repo. Northern-hemisphere means render identically before and after.
- **State.** CI green, merges cleanly into master (41 commits behind, no overlap in these functions).

### Added in 82aac0cf

Nothing in `pmagpy/test` exercised these functions, so I added `pmagpy/test/test_ipmag_compilation_kent.py` (9 tests): n honored below and above the compilation length, a user-supplied compilation, the Kent mean staying in the pole's hemisphere, `PIERCE2022_F_COMPILATION` length and range, and the ellipse hemisphere at the `plot_ell` and `plot_pole_ellipse` levels (including that the two `lower` choices are antipodal point sets). Six of the nine fail against master. Map-dependent tests skip when cartopy is absent, which is the case for this branch's CI config (master installs cartopy since #866). Full active suite: 784 passed.

### Suggestions (non-blocking)

1. `lower=False` is a public default change on `plot_pole_ellipse`. Only notebooks that passed a southern Kent mean and got an antipodal ellipse will see a difference, but worth a line in the description and a check of the Essentials JupyterBook for such calls.
2. The description still frames the change as a reduction from `n=10000`; the comment above has the accurate account (the old loop never used `n`, so `n=1000` is a tenfold increase in work). Folding that into the body would leave a cleaner record.

Generated by Claude Fable 5.1 (Claude Code) on behalf of NSH

PR #904, opened 2026-08-04
Versioning issue pmagpy vs pmagpy-cli

apivarunas:
# Summary
This PR fixes a packaging/versioning issue that leaves pmagpy-cli and pmagpy out of sync during installation. This, I believe, is what is causing the whack-a-mole problems in #769. The CLI package previously depended on pmagpy without pinning the exact matching version, which allowed pip to resolve a different compatible release and could lead to environments where the GUI/CLI entry points and the installed core library no longer matched. I spend some time running this through Copilot and here is what was come up with.

## What changed
- pmagpy-cli now requires the exact matching pmagpy version derived from the shared version string.

- A regression test was added to ensure this dependency contract remains in place.

## Steps to make the fix effective
To actually resolve the issue for users, pmagpy-cli needs to be re-released with this updated packaging metadata. In practice, the release flow should be:

- Publish or update pmagpy with the intended version.
- Publish pmagpy-cli from the same code state with the matching version pin.
- Ensure the two packages are released in sync so pip installs resolve the correct compatible pair.

## How to avoid this in the future
To prevent this from happening again, the release process should treat pmagpy and pmagpy-cli as coupled packages:

- release them together from the same tag or release commit,
- keep the versioning scheme aligned,
- and avoid loose dependency relationships between them.

## Long-term recommendation
The longer-term goal should be to simplify the packaging model by consolidating the core library and GUI/CLI entry points into a single installable package. If the split must remain, pmagpy-cli should continue to function as a thin wrapper released in lockstep with pmagpy and pinned to the exact matching version.

### DIFF
--- a/command_line_setup.py
+++ b/command_line_setup.py
@@ -2,8 +2,7 @@ from setuptools import setup, find_packages
 from pmagpy import version
 from programs_list import programs_list
 
-version_num = version.version.strip('pmagpy-')
-shared_version_requirement = f'pmagpy=={version_num}'
+version_num = version.version.removeprefix('pmagpy-')
 
 # skip the html header at the top of README.md
 with open('README.md', encoding='utf-8') as f:
--- a/command_line_setup.py
+++ b/command_line_setup.py
@@ -3,6 +3,7 @@ from pmagpy import version
 from programs_list import programs_list
 
 version_num = version.version.strip('pmagpy-')
+shared_version_requirement = f'pmagpy=={version_num}'
 
 # skip the html header at the top of README.md
 with open('README.md', encoding='utf-8') as f:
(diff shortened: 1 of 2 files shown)


Swanson-Hysell:
I am guilty of putting together pip releases for pmagpy without doing parallel ones for pmagpy-cli. It makes sense to be cleaner about it and do them together.

apivarunas:
> I am guilty of putting together pip releases for pmagpy without doing parallel ones for pmagpy-cli. It makes sense to be cleaner about it and do them together.

All good! As has been discussed, future packaging and versioning might mean that the parallelization and issues thereof might go away.

apivarunas:
Okay, review looks like it caught a few future issues and the test passes and the checks pass. Mergin'!

PR #905, opened 2026-08-15
Defunct scipy replacement

apivarunas:
The functions `scipy.mean` and `scipy.array` have been moved over to numpy as of some time ago. These were lingering in the generic converter, which I ran into because I was trying to do some generic imports.


Swanson-Hysell:
Thanks for catching this. I just confirmed the bug and the fix so will go ahead and merge.

PR #911, opened 2026-08-27
Follow-up to the curie.py replacement 

Swanson-Hysell:
Follow-up to the curie.py replacement (#910)

Makes some changes building on #910. Addressing a couple issues found by executing the notebooks and running the CLI against curie_example.dat:

- Plots and MagIC notebooks: Put in import using `rmag`.
- curie.py: smooth the signal before estimating, matching `curie_temperature_estimates`
- Tests: data checks on curie_example.dat
- PmagPy-cli.ipynb: point at the new `curie_<method>.<fmt>` output.

The `-w` default change differs from the (unreleased) #910 CLI; flags are compatible with the legacy curie.py.


apivarunas (approved):
This fixes some missed issues with #910, principally in the realm of the notebooks, making notebook imports consistent (and adding some missing ones) as well as adjusting verbiage and runs. I think that the added testing is a little messy (I think Copilot recommended some similar monkey-patched approach, but I decided against it), but if it adds some crucial information that's fine.

Swanson-Hysell:
Thanks, Anthony. I think that the test that swaps in fake estimators is likely worth pulling out. The tests that run on the actual example data are the ones that, I think, makes sense to keep.

Issue #836, opened 2026-03-08
irm_unmix doesn't play nicely in notebook


Swanson-Hysell:
**What command was run:**

When 
```
import programs.irm_unmix as irm_unmix
```
is run in the PmagPy_plots_analysis.ipynb notebook there can be plotting backend errors.

`irm_unmix.py` hardcodes `matplotlib.use('Qt5Agg')` and imports PyQt5 at the top level because it's a standalone GUI program. The real problem is that irm_unmix wasn't designed to be imported as a library — it's a Qt GUI application. 

It likely makes sense to remove this notebook approach to using it. This approach is aligned with such functionality being within rockmagpy notebooks.

Issue #793, opened 2025-10-08
second ipmag.plot_dmag() demo in PmagPy_plots_analysis.ipynb not resulting in plot


Swanson-Hysell:
**What command was run:**

In the `PmagPy_plots_analysis.ipynb` notebook there are multiple issues with the code block:

```
ipmag.download_magic("magic_contribution_16533.txt", dir_path="data_files/download_magic", 
                     input_dir_path="data_files/download_magic")
status,data=cb.add_sites_to_meas_table('data_files/download_magic')
df=data[data.method_codes.str.contains('LT-T-Z')] # select the thermal demag data
df=df[['specimen','treat_temp','magn_moment','quality']]
df=df.dropna(subset=['treat_temp','magn_moment'])
ipmag.plot_dmag(data=df,title="Thermal demag",fignum=1, dmag_key='treat_temp')
```

The first is that data_files/download_magic does not exist with the repository. 

The second is that when that directory is created and that file put in there, no plot is produced when the code is run.

Issue #781, opened 2025-10-07
pmagplotlib.plot_init() NonGuiException


Swanson-Hysell:
**What command was run:**

`ipmag.histplot(data=norm, xlab='Gaussian Deviates', save_plots=False, norm=-1)`
within `PmagPy_calculations.ipynb`

**What happened:**

`NonGuiException error`

`NonGuiException: Matplotlib is currently using module://matplotlib_inline.backend_inline, which is a non-GUI backend, so cannot show the figure.`

It looks like `pmagplotlib.plot_init()` is not able to figure out that the code is being run in a notebook.

```
---------------------------------------------------------------------------
NonGuiException                           Traceback (most recent call last)
Input [In [72]](vscode-notebook-cell:?execution_count=72), in <cell line: 2>()
      1 # alternatively we can plot with ipmag.histplot:
----> [2](vscode-notebook-cell:?execution_count=72&line=2) ipmag.histplot(data=norm, xlab='Gaussian Deviates', save_plots=False, norm=-1)

File ~/0000_Github/PmagPy/pmagpy/ipmag.py:15422, in histplot(infile, data, outfile, xlab, binsize, norm, fmt, save_plots, interactive)
  15420     print('-W- Not enough data found')
  15421     return False, []
> 15422 fig = pmagplotlib.plot_init(1, 8, 7)
  15423 try:
  15424     len(D)

File ~/0000_Github/PmagPy/pmagpy/pmagplotlib.py:139, in plot_init(fignum, w, h)
    137 fig = plt.figure(num=fignum, figsize=(w, h), dpi=dpi, clear=True)
    138 if (not isServer) and (not set_env.IS_NOTEBOOK):
--> [139](https://file+.vscode-resource.vscode-cdn.net/Users/unimos/0000_Github/PmagPy/~/0000_Github/PmagPy/pmagpy/pmagplotlib.py:139)     plt.get_current_fig_manager().show()
    140     # plt.get_current_fig_manager().window.wm_geometry('+%d+%d' %
    141     # (fig_x_pos,fig_y_pos)) # this only works with matplotlib.use('TKAgg')
    142     fig_x_pos = fig_x_pos + dpi * (w) + 25

File ~/opt/anaconda3/lib/python3.9/site-packages/matplotlib/backend_bases.py:2815, in FigureManagerBase.show(self)
   2809 if sys.platform == "linux" and not os.environ.get("DISPLAY"):
[...]

PR #919, opened 2026-09-19
Stop programs.* imports from clobbering the matplotlib backend (fixes #909)

### LINKED ISSUE #909: Importing programs.* modules breaks inline plotting in notebooks
A bunch of the modules under `programs/` call `matplotlib.use("TKAgg")` or `("WXAgg")` when imported
import time. The guards only check whether the backend is already a GUI one:

As a result importing them from a notebook switches the backend away from inline and figures stop appearing:

%matplotlib inline
ipmag.histplot(...)          # backend: inline   -> figure renders
import programs.common_mean  # backend: TKAgg
ipmag.histplot(...)          #                   -> no figure

I think that this is the cause of issues #836 and part of #793, where %matplotlib inline workaround is dealing with the issue of an earlier import clobbering the backend. 

I think that we should be able to skip the guards when pmag_env.set_env.IS_NOTEBOOK is set, or
when the backend is already non-GUI. Setting a GUI backend is what should be done when the modules run as CLI programs.

Swanson-Hysell:
Fixes #909.

## Problem

Modules under `programs/` called `matplotlib.use("TKAgg")` (or `"WXAgg"`) at import time whenever the backend was not already that GUI backend. Importing one of them from a notebook replaced the inline backend and figures stopped appearing:

```python
%matplotlib inline
ipmag.histplot(...)          # renders
import programs.common_mean  # backend is now TKAgg
ipmag.histplot(...)          # nothing
```

While fixing this a second path turned up: `programs/conversion_scripts/tdt_magic.py` (imported eagerly by the package `__init__`) and `livdb_magic.py` forced `WXAgg` unconditionally, so `from programs.conversion_scripts import cit_magic` had the same effect.

This is the root cause behind the `%matplotlib inline` workarounds added for #836 and #793.

## Fix

- New `set_env.set_backend_if_unset(preferred)` in `pmag_env/set_env.py`. It requests a GUI backend only when matplotlib has not yet resolved one, and never inside a notebook. A backend chosen earlier is respected: the inline backend that Jupyter kernels set through `MPLBACKEND`, an explicit `matplotlib.use()`, a test suite's `Agg`. A companion `backend_already_chosen()` inspects matplotlib's auto-select sentinel, with a fallback for older releases.
- All import-time backend selection under `programs/` now goes through the helper (54 modules with the old `get_backend() != ...` guard, the two conversion scripts, `forc_diagram.py`, and `programs/__init__.py`, whose old guard also triggered backend resolution by calling `get_backend()` first).
- In `histplot.py` and `quick_hyst.py` the selection ran after `pmagpy` imports. It now runs before them.
- The four wxPython GUIs (`demag_gui`, `pmag_gui`, `magic_gui`, `thellier_gui`) still force `WXAgg`; they embed matplotlib in wx windows and are not meant to be imported into notebooks.
- `ani_depthplot2`, `tdt_magic` and `livdb_magic` are added to `programs/program_envs.py` so the package init selects `WXAgg` for them. On master their own `WXAgg` request won by forcing the backend after the package init had picked `TKAgg`; with the helper that request is a no-op, so without the entry they ended on `TKAgg`. The other conversion scripts, which got `WXAgg` on master only through the eager package import of `tdt_magic`, never plot and keep the `TKAgg` default.

Command-line behaviour is unchanged. Checked from subprocesses: a plain program import still selects `TKAgg`, `core_depthplot` still gets `WXAgg` and `make_magic_plots` still gets `Agg` from `program_envs`, and `MPLBACKEND=Agg` is respected.
- Programs started from a notebook cell with `!eqarea.py ...` keep working. ipykernel exports `MPLBACKEND=module://matplotlib_inline.backend_inline` and every subprocess inherits it, so with the first version of this branch such a program kept the inline backend and `plt.show()` printed `Figure(640x480)` instead of opening a window. `set_backend_if_unset` now overrides a chosen backend in exactly one case: it is an inline backend and IPython is not present in the process, where inline cannot render anything. A notebook front end with an unrecognised shell class (#781) still has IPython in the process, so its inline backend is left alone.

## Tests

New `pmagpy/test/test_programs_backend.py`:

- unit tests for `set_backend_if_unset` (already chosen, nothing chosen, notebook)
- import tests for several `programs.*` modules under `Agg`, including `programs.conversion_scripts.cit_magic`
- a guard that fails if a module-level `matplotlib.use()` or the old `get_backend()` pattern reappears under `programs/` outside the GUI allowlist
- an end-to-end test that starts a real `ipykernel`, imports `programs.common_mean`, `programs.eqarea_magic` and a conversion script, and asserts the inline backend survives and a figure is emitted. CI installs `jupyter`, so this runs there.

- unit tests for the inherited-inline override, the unrecognised-IPython-shell case, a non-inline `MPLBACKEND`, and `is_inline_backend`
- a subprocess test that reproduces the `!program` child process directly, and the kernel test also spawns a child from inside the kernel and asserts it gets `TkAgg`

- a test that runs every `program_envs` entry as a script and checks the backend the package init selects

Full `pytest pmagpy/test`: 867 passed, 20 skipped in an environment without wxPython (as on CI); 873 passed, 14 skipped with wxPython.

## Follow-ups (not in this PR)

- `programs/foldtest.py` calls `main()` at module level with no `__name__` guard, so importing it prints the usage and exits.
- The `%matplotlib inline` cells added for #836 and #793 can now be dropped.
- `programs/conversion_scripts/__init__.py` imports `tdt_magic` eagerly, and `tdt_magic` and `livdb_magic` define `wx.Frame` classes at module level, so the whole package is unimportable without wxPython.

Generated by Claude Fable 5.1 (Claude Code) on behalf of NSH


Swanson-Hysell:
The failing `build (3.12)` job was `TestNotebookInlinePlotting::test_importing_program_keeps_inline_backend`. The kernel test imported `programs.conversion_scripts.cit_magic`; the package `__init__` imports `tdt_magic` eagerly, and `tdt_magic` imports `wx` at module level, which CI does not install. This is a pre-existing limitation of the conversion_scripts package rather than something this branch introduced (the in-process import test already skipped in that case). 266128d8 makes the kernel test include that import only when wxPython is importable.

Reviewing the branch against every touched program also turned up a regression, fixed in 182b3639: ipykernel exports `MPLBACKEND=module://matplotlib_inline.backend_inline` to its subprocesses, so a program run from a notebook cell with `!eqarea.py ...` kept the inline backend and `plt.show()` printed `Figure(640x480)` instead of opening a window. `set_backend_if_unset` now replaces an inline backend when no IPython is present in the process, which is the only case where inline cannot render. The PR description is updated accordingly.

Checked in all 58 modules that the helper is called before any `pmagpy`/pyplot import, so it never becomes a no-op because the backend was already resolved; no module-level backend forcing remains under `programs/` outside the four wx GUIs.

Generated by Claude Fable 5.1 (Claude Code) on behalf of NSH

Swanson-Hysell:
This PR touches a lot of the programs in addition to the notebooks so I am going to keep testing this branch for a couple days to make sure that there aren't unintended consequences that cause issues within the programs or notebooks.

Swanson-Hysell:
Local verification beyond the unit tests, all in the `pmagpy-dev` environment (matplotlib 3.10.8, wxPython, cartopy, ipykernel), comparing this branch against `origin/master` in separate repo copies:

**Terminal import sweep, all 207 `programs` modules.** Each module imported in a fresh interpreter with `sys.argv[0]` set to its program name, four ways: master and branch, each with and without an inherited `MPLBACKEND=module://matplotlib_inline.backend_inline`. 203/207 import cleanly in every variant; the same four fail identically on master (missing `habanero`, two deprecated basemap modules, `run_multi_samples`). Under the inherited inline variable the branch ends on exactly the same backend as in a plain terminal for all 207 modules. The sweep is what turned up the 29 modules that had lost `WXAgg`; after 1ef56c29 the three that matter (`ani_depthplot2`, `tdt_magic`, `livdb_magic`) get `WXAgg` again.

**Kernel import sweep.** In one real ipykernel: `ipmag.histplot` → `import programs.common_mean` → `ipmag.histplot` emits 1 → 1 figures on the branch (backend stays inline) versus 1 → 0 on master (backend becomes `TKAgg`). Then all 203 non-GUI modules imported in sequence; the inline backend survives every one, and a final `plt.show()` plus `ipmag.histplot` produce 2 `display_data` messages.

**CLI corpus from PmagPy-cli.ipynb.** The 157 `!program ...` cells were extracted and 142 run against master, branch, and branch with the inherited inline variable (skipping `cat`/`mv`/network programs). Exit codes are identical across all three runs. Every output difference is bootstrap randomness (`bootams`, `fisher`, `fishrot`, `foldtest`, `uniform`, `watsons_v`, `revtest`, `tk03`, `find_ei`) or a traceback line number shifted by one by the edit. Fourteen commands fail with the same error on master (numpy float alias removal in `di_rot`/`di_eq`/`di_tilt`/`vector_mean`/`common_mean`, pandas `delim_whitespace` in `scalc`, `aarm_magic` signature, `gaussian`, `aniso_magic`); those are pre-existing and separate from this PR.

**Headless notebook re-execution (nbclient).** `PmagPy_plots_analysis.ipynb`: master keeps 109 of 227 images and loses every figure from cell 156 (`from programs.forc_diagram import *`) onward, which is #909 at scale; the branch keeps 219 of 227, and the eight missing are two cells broken by pandas 3 dropping `delim_whitespace` and one map cell that is empty on master as well. `PmagPy_calculations.ipynb` (29 → 29) and `PmagPy_MagIC.ipynb` (167 → 167) are unchanged. `SVEI_demo.ipynb` and `PmagPy_online.ipynb` lose figures only in cells that fail on missing downloaded data or the absent `imageio` package.

Generated by Claude Fable 5.1 (Claude Code) on behalf of NSH

Issue #360, opened 2018-05-29
Populating Demag GUI measurement boundary boxes


Caoimhinmg:
As Jeff Gee suggests:

> For example, when I load a new set of data (i.e. in a project folder with no previous interpretations) the drop down menus for the upper and lower bounds are both empty (first figure). So I can’t select bounds before adding a fit. When I add a fit, it calculates a pca for the total range of data by default (second figure). Thus my suggestion to just skip a step and have it populate a starting fit and populate the bounds drop down menus.

In other words Demag GUI should have populated measurement boundary boxes even before a fit is made considering it already adds a fit when you pick bounds in the list or plot but you can't use the drop down menu.

apivarunas:
This issue has now been covered with PR #920. Closing!

PR #920, opened 2026-09-19
Issue #360 fix

apivarunas:
This PR addresses issue #360.

1. Dropdown measurement step boxes are now populated even without a fit.
2. If steps are selected, then the button "Add Fit" will fit those selected steps.
3. If no steps are selected, then the button "Add Fit" will fit the first and last steps, as before.

Tested by running local version of `demag_gui` which confirms that this behaves as expected.


Swanson-Hysell:
This is great. The one behavior that is a bit odd is that whenever an already made fit is selected, the dropdowns show its bounds, so `on_btn_add_fit` reads them and so a new second fit on the specimen is an exact copy of the current one. Maybe that is ok, but another approach could be to have the behavior be like the previous behavior where a new fit that is added when a current fit is selected spans that full range like before. 

This would likely do it:
```
fmin = fmax = None
if self.current_fit is None:
    # bounds picked in the dropdowns before any fit exists (issue #360)
    fmin = str(self.tmin_box.GetValue()) or None
    fmax = str(self.tmax_box.GetValue()) or None
```

Another thing to consider here is that selecting bounds with no fit does nothing until Add Fit is pressed. Jeff Gee's suggestion in the issue was to populate a starting fit. Clicking in the measurement list already auto-creates a fit when none exists, so calling `on_btn_add_fit` from `get_new_PCA_parameters` when both boxes are set and `current_fit` is None would make the dropdowns behave the same way.

apivarunas:
Good suggestions @Swanson-Hysell. I think these 2 more edits have fixed the two things you've brought up. Doesn't seem to clone the fit anymore, and once both bounds are selected a fit is auto-picked (and once that fit is made, it doesn't repeat that process).

Swanson-Hysell:
Just did some testing with the latest version — looks good!

